Issue #111, opened 2014-09-15
Group write permission for the `translations` dir


asprega:
I was wondering why, when dumping the translations, the `translations` dir doesn't get the g+w permission.
In other words: if I dump the translation as a given user, then I try to do it again as a different user, I get the following error:

`[Symfony\Component\Filesystem\Exception\IOException]      
  Failed to remove file "web/js/translations/messages/en.js".`

This could be useful in at least two multi-user situations I can think of:
a) a shared doc-root, where multiple developers would want to run commands
b) a non-shared doc-root which is used e.g. by a developer and a jenkins instance

Is this a requirement that can be accepted? If so, would you hardcode the new permission bitmask in code or make it configurable? I could create a pull request for that.

asprega:
I might be wrong here, but based on this line of code in the `TranslationDumper`: `$this->filesystem->mkdir(dirname($file), 0755);` I don't think I can do anything with the umask. If it was 0777 I could have used the umask to tune the permissions. What about updating that value to `0777`? Or do you have any other suggestions other than hacking up a console command to fix permissions?

stof:
@willdurand the bundle should indeed use 0777 to let the umask control it. If it uses 0775 or 0755, it controls it itself instead of letting the system control it.

willdurand:
@asprega @stof I pushed a quick fix, let me know.

asprega:
Great! I found another two occurrences where the 0755 mode should be removed imho:

https://github.com/willdurand/BazingaJsTranslationBundle/blob/9064423df5c13aa989effdced01264ed3d2a24cb/Dumper/TranslationDumper.php#L185

https://github.com/willdurand/BazingaJsTranslationBundle/blob/61b8c50ed78f20fc592525b2ad1ecee99060fd92/Tests/Dumper/TranslationDumperTest.php#L27 (this one is not necessary but may be good for consistency)

What do you think?

willdurand:
@asprega would you mind create a PR? I think it would be better to remove all occurences.

asprega:
sure!

asprega:
Thanks for merging! Looking forward to a tagged release including this.

PR #129, opened 2015-01-03
Now using chmod 0777 on all occurrences of mkdir.

asprega:
See issue #111.

I did not write any tests on the change since it doesn't seem like they would add any value (besides the fact that it would not be trivial to verify since the default dir permissions would change based on the umask of the user running the tests). Please let me know if you think we need them.

### DIFF
--- a/Dumper/TranslationDumper.php
+++ b/Dumper/TranslationDumper.php
@@ -182,7 +182,7 @@ class TranslationDumper
                         ))
                     );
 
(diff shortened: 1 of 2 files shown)


willdurand:
I think it's fine as is.

willdurand:
thank you very much!

asprega:
No problem, thanks for merging! Looking forward to a tagged release including this.

PR #133, opened 2015-02-19
Change all private variables and methods to protected

TomiS:
I noticed the need for this when I tried to inherit this bundle and extend the controller with my own. Having protected variables would make this easier and cleaner.

### DIFF
--- a/Controller/Controller.php
+++ b/Controller/Controller.php
@@ -20,42 +20,42 @@ class Controller
     /**
      * @var TranslatorInterface
      */
-    private $translator;
+    protected $translator;
 
     /**
      * @var EngineInterface
      */
-    private $engine;
+    protected $engine;
 
     /**
      * @var TranslationFinder
      */
-    private $translationFinder;
+    protected $translationFinder;
 
     /**
      * @var array
      */
-    private $loaders = array();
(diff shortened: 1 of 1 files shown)


stof:
changing everything to protected means that any change to one of these property or method (including not using it anymore for the job) is a BC break.

TomiS:
Hmm, good point. Which kind of brings me to another point. Is it really good practice to make controllers this heavy with logic? Personally I might have done all the magic in a separate service class and injected all those config variables there (that are now controller constructor parameters). In that case the controller would have remained without constructor parameters and it would have been easier to overload. Just wondering if there's a benefit in the strategy chosen in this bundle.

stof:
@TomiS what is your use case for overwriting it ? This could help identifying the missing extension points

TomiS:
@stof Well, I wanted to experience with different output formats but also wanted to add apidoc annotations using NelmioApidocBundle.

TomiS:
Most importantly I'm trying to get all translation domains for all languages on one request in json format. But I'm not sure if the original api is providing that option.

stof:
all languages at once is possible. However, the current controller is indeed limited to a single translation domain.

PR #138, opened 2015-06-08
Added note about Assetic

Rvanlaak:
... because a manual dump is needed to refresh the static translation files.


willdurand:
thanks!

PR #139, opened 2015-06-08
Add /js as routing prefix

rvanlaak:
... so controller will take over if translation assets are not dumped yet. That way the url is consistent with the Assetic configuration.

Or maybe the Assetic and the dump path should be changed?

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -38,6 +38,7 @@ Register the routing in `app/config/routing.yml`:
 # app/config/routing.yml
 _bazinga_jstranslation:
     resource: "@BazingaJsTranslationBundle/Resources/config/routing/routing.yml"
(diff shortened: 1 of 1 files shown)


stof:
This route has nothing to do with Assetic at all.

Rvanlaak:
You are right @stof , the documentation about Assetic mentiones `/js` as base path, so it's about the dump command.  

https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/doc/index.md#assetic

Correct me if I'm wrong, but by adding `/js` as a prefix the controller will take over in case the translations are not dumped yet.

The bundle now uses two paths, `/translation` for the usual urls and `/js/translation` for the dump command.

rvanlaak:
No response, assume this will not get merged.

PR #142, opened 2015-06-14
Replace string casting on ConfigCache instance with getPath() to omit…

akucherenko:
With release of Symfony 2.7 deprecated warnings were added:

DEPRECATED - ConfigCache::__toString() is deprecated since version 2.7 and will be removed in 3.0. Use the getPath() method instead.

See https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Config/ConfigCache.php#L49
This PR fixes warning when casting instance of ConfigCache to string.

### DIFF
--- a/Controller/Controller.php
+++ b/Controller/Controller.php
@@ -164,10 +164,16 @@ class Controller
             }
         }
 
+        if (method_exists($cache, 'getPath')) {
+            $cachePath = $cache->getPath();
+        } else {
+            $cachePath = (string) $cache;
+        }
+
         $expirationTime = new \DateTime();
         $expirationTime->modify('+' . $this->httpCacheTime . ' seconds');
         $response = new Response(
-            file_get_contents((string) $cache),
+            file_get_contents($cachePath),
             200,
             array('Content-Type' => $request->getMimeType($_format))
         );


willdurand:
Oops, looks like your patch broke the test suite...

akucherenko:
This patch requires Symfony 2.7 (ConfigCache->getPath() was added), while travis tests only up to 2.6. There can be added some workaround or just update dependencies in composer.json

willdurand:
Thing is, we can't only provide for 2.7. It must work with 2.3+ I'd say.

akucherenko:
You can keep 2.3-compatible code in separate tag and rename current branch-alias for master to 2.7-dev. It won't break existing usage of your bundle since composer require it like: `"willdurand/js-translation-bundle": "2.3.*@dev"`. Does it work for you?

willdurand:
Nope. I'd rather add a `method_exists()` call or something

akucherenko:
I've used version_compare() if you don't mind. It just looks more accurate for me.

Controller/Controller.php
```diff
@@ -164,10 +165,16 @@ public function getTranslationsAction(Request $request, $domain, $_format)
             }
         }
 
+        if (version_compare(Kernel::VERSION, '2.7.0', '>=')) {
```
stof: This is a very bad idea. Someone using components directly can have a different version for the Config component and the Kernel. You need to detect whether the Config component supports the new API, not what the kernel version is

akucherenko: @stof Thanks for this point, didn't thought about that. So now it is method_exists().

stof:
:+1:

akucherenko:
@willdurand Did you check it? Is it OK?

willdurand:
yup!

willdurand:
thank you

PR #143, opened 2015-06-26
Translator > Added support for language fallback when given a language + culture

ruudk:
Load your translations like this:
`translations?locales=nl_NL,nl_BE,nl`

When translating it will first search for the full ISO code `nl_BE`. If it can't find a translation it will fallback to the language `nl`.


ruudk:
@willdurand How can I create the `translator.min.js` file? Maybe it's good to have a doc for that?

willdurand:
Using YUICompressor I'd say :)

ruudk:
What do you think of the rest? Is it mergable? :)

ruudk:
Added the minified translator.js

Resources/js/translator.js
```diff
@@ -53,6 +53,17 @@ var Translator = (function(document, undefined) {
         var _locale = locale || currentLocale || localeFallback,
             _domain = domain;
 
+        /**
+         * If the _locale looks like nl_NL or nl-NL
+         * Try the language as the locale
+         */
+        if (undefined == _messages[_locale]) {
+            var tryLanguage = _locale.match(/^([a-z]{2})(_|-)[A-Z]{2}$/)
```
stof: Symfony actually support any number of sublocales, not only one.

And you should not restrict it to 2 letters IMO (especially not enforcing the case), as it is inconsistent with Symfony.

ruudk: So what should I do here?

stof: I suggest looking at the logic used in Symfony and porting it to JS. The goal is to have it behave the same than in PHP, so it would make sense

stof: and btw, the ``localeFallback`` could be ``en_US`` too, which should itself fallback to ``en`` to mimic the Symfony behavior

ruudk:
This is already fixed by another PR

PR #144, opened 2015-07-03
Safer twig locale

foaly-nr1:
The recommended way of storing locales is language code underscore country code, e.g. en_GB.

> The term locale refers roughly to the user's language and country. It can be any string that your application uses to manage translations and other format differences (e.g. currency format). The ISO 639-1 language code, an underscore (_), then the ISO 3166-1 alpha-2 country code (e.g. fr_FR for French/France) is recommended.
> – http://symfony.com/doc/current/book/translation.html

Therefore, a safe way of using the locale in the lang attribute would be:

```twig
<html lang="{{ app.request.locale|split('_')[0] }}">
```

This works with both an **en_GB** locale and an **en** locale.

Source: http://stackoverflow.com/a/30538515/2106834

### DIFF
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -58,7 +58,7 @@ Then add the current application's locale into your layout, by adding a `lang`
 attribute to the `html` tag:
 
 ```html
-<html lang="{{ app.request.locale }}">
+<html lang="{{ app.request.locale|split('_')[0] }}">
 ```
 
 Now, you are done with the basic setup, and you can specify the [translation


willdurand:
thanks

marcosdsanchez:
The problem with that approach is that it won't return the country code. It will only return the language. If you want a full locale as supported in http://www.w3.org/TR/html401/struct/dirlang.html#h-8.1.1 . You should do this instead: 

```
<html lang="{{ app.request.locale|replace('_', '-') }}">
```

foaly-nr1:
Problem is, BazingaJsTranslationBundle won't recognise it anymore. We need full support for locales first.

PR #146, opened 2015-08-06
Translator > Added support for other languages from symfony intl bundle

zocimek:
Using language es_419 is currently not supported due to filtering the locales. I've tested regexp against Intl::getLocaleBundle()->getLocales() and added also some test cases.


stof:
What aboout using the same regex than the one used in the Symfony Translator to restrict valid locales ? https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Translation/Translator.php#L484

stof:
@zocimek FYI, you committed with an email address which is not associated to your github account, so Github does not match these contributions as being ours (in the contributor graph for instance)

zocimek:
@stof thanks for your advice I've changed the pattern to be valid with symfony.

zocimek:
@stof @willdurand Can anyone merge this change ?

stof:
Given that you accept dashes now, you should normalize them to underscores, to match the Symfony behavior

stof:
and I cannot merge this. I'm not a collaborator on the repo

zocimek:
@stof you mean something like this ? 

```php
$locales = array_unique(array_map(function ($locale) {
    if (strrchr($locale, '_') !== false) {
        $locale = substr($locale, 0, -strlen(strrchr($locale, '_')));
    }
    
    return trim($locale);
}, $locales));
```

fatalcoder:
@willdurand @monteiro Could you merge this change?

zaytsevfuu:
Would solve this issue: https://github.com/willdurand/BazingaJsTranslationBundle/issues/188

monteiro:

Issue #135, opened 2015-04-27
Locales with region are not supported


Agares:
Hey, when I have my html lang attribute defined as "de-DE", which is correct, according to http://www.ietf.org/rfc/bcp/bcp47.txt and import translations for "de" (<script src="{{ url('bazinga_jstranslation_js') }}?locales=de"></script>), then Translator.trans('MyMessage') returns message in default language (I expect it to return them in german).

danbettles:
Hello there

I'm facing a similar problem.  

Our app allows "en_GB" and "en".  We support other flavours of English so that dates and money values, for example, are correctly localized.  To make things easier, all English translations are currently stored in "en" files.

I'm running the latest version of the bundle.  When I request "en_GB" translations, I don't get the "en" translations.  By the way, I'm expecting the "en" translations only because that's what Symfony would give me.  Even with the fallback locale set to "en", I don't get the "en" translations.

Is that expectation unreasonable: should I expect to get the "en" translations if there aren't any "en_GB" translations?  (I mean: it may be undesirable get something other than specifically what you asked for.)  One way or another, I'll fork the code and see if I can find a sensible solution for us.  I'll perhaps issue a PR after.

Thanks very much for the bundle, by the way.  We've found it really useful and are happily using it through RequireJS.

Best,
Dan

willdurand:
I think it only works well the other way around for now. How does Symfony translator behave?

Agares:
Symfony translators looks first for messages.de_DE.yml, then messages.de.yml.

willdurand:
For ˋde_DEˋ?

Agares:
Yes, because it is the format that Symfony supports for its locale. Valid "lang" attribute value for <html /> tag is "de-DE".

willdurand:
Then lets add this behavior.

willdurand:
I think adding this behavior should not be too much work as the `Translator` is well-tested. Anyone willing to help on this?

danbettles:
Afternoon, both.

I've just pushed some changes to my fork of the repository.  See https://github.com/powderblue/BazingaJsTranslationBundle.  I tried a couple of approaches before going with this - what I think is fairly un-instrusive.  I don't doubt, however, that things could be improved and I realise this is only a part of the solution.

*So far*, these changes appear to have solved my problem: when my app is operating in the "en_GB" locale I will get translations for "en" if translations for the requested locale are not available.

Dan

Agares:
[...]

PR #147, opened 2015-08-10
Locales with region are not supported

danbettles:
Hello

This pull request deals with https://github.com/willdurand/BazingaJsTranslationBundle/issues/135.  I'm sorry it's taken so long to get back to you.  I've been working on a different project all this time.

Please let me know if the code's of any use and I'll bring it up to date.

Dan


.gitignore
```diff
@@ -1,2 +1,3 @@
-vendor/
-composer.lock
+vendor/
+composer.lock
+/nbproject/
```
stof: nbproject should not be there. You should ignore it locally instead as it is specific to your setup: https://help.github.com/articles/ignoring-files/

stof:
File permissions must not be changed to make them executable.

If you are on Windows, here is how to fix that:

- tell git to ignore filemode changes (as Windows does not support file mode anyway): ``git config --global core.fileMode false``. Clever installers for git on windows are doing it by default, but not all of them are clever.
-  undo the file mode change for all files changed in this PR, using the low-level git API (as there is no system tool to change the file mode): ``git update-index --chmod=-x <path>`` (repeat it for each changed file)

stof:
And this code is flawed. Symfony does not restrict to have a single level of fallback. see https://github.com/willdurand/BazingaJsTranslationBundle/pull/146 which is allowing all valid symfony locales

PR #150, opened 2015-09-16
Widen TranslationFinder::getLocations() visibility to use is in an in…

tgabi333:
…teritant class

Our use case: we like to add new locations. And now we have to override almost the whole class, instead it would be better if we can inherit from it and just append some lines to its result.


willdurand:
thanks

PR #156, opened 2015-11-10
Allow Symfony 3.

aitboudad:


composer.json
```diff
@@ -11,15 +11,15 @@
...
+        "symfony/console": "~2.3|~3.0",
+        "symfony/intl": "~2.3|~3.0"
     },
     "require-dev": {
-        "symfony/yaml": "~2.3",
-        "symfony/browser-kit": "~2.3",
-        "symfony/twig-bundle": "~2.3"
+        "symfony/yaml": "~2.3|~3.0",
+        "symfony/browser-kit": "~2.3|~3.0",
+        "symfony/twig-bundle": "~2.3|~3.0"
```
stof: you should add the dev requirement on the phpunit bridge to catch deprecations

stof:
you are missing the installation of dependencies

composer.json
aitboudad: done

willdurand:
thanks!

.travis.yml
```diff
@@ -17,15 +25,11 @@ script:
     - phantomjs Resources/js/run-qunit.js file://`pwd`/Resources/js/index-with-es5-shim.html
 
 matrix:
-    allow_failures:
-        - env: SYMFONY_VERSION=dev-master
-        - php: hhvm
+    fast_finish: true
     include:
+      - php: 5.3.3
+        env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
```
stof: this does not work, as you don't use this variable

willdurand: ah. I suppose it is not an env var Composer would read by itself then?

willdurand: those options are available on the `update` command only anyway, so removing them.

Issue #151, opened 2015-09-22
Lack of support for custom translator


lchachurski:
JS translations seems to ignore custom Translator class. What I'm trying to do, is to display translation keys instead of translations to ease translation process. It works fine for standard translations, but not at all for JS ones. Setting defaultDomain to '' didn't help as well. Any hints?

``` yml
parameters:
    translator.class: Acme\HelloBundle\Translation\Translator
```

source: http://symfony.com/doc/current/cookbook/bundles/override.html#services-configuration

herodrigues:
Have you tried to set the default domain to messages and put your translation files inside the Resources/translations folder of your bundle?

chrifmarwen:
I have the same problem.

stof:
I don't understand how the JS should use your custom class. This bundle does not use the Translator class at all (as it reimplements it in JS). The PHP side only uses the translation loading.

If you want to display translation keys rather than messages, you will need to overwrite the JS ``Translator`` variable with your own implementation.

chrifmarwen:
There is acommand line that uses the translator finder to generate JS translation. So whenever our translation are defined it will find them using the config.yml.

stof:
@chrifmarwen I don't understand what you mean. The JS translation files are generated by loading the translation catalogue with the loaders and dumping it as JSON or JS. It does *not* call the Translator.

lchachurski:
Closing this one as @stof has a good point here.

PR #159, opened 2016-02-12
Ability to using custom translator path

chrifmarwen:
based on : https://github.com/willdurand/BazingaJsTranslationBundle/issues/151
i added the ability to get translation message based on the custom translator path.
getting this paths from the config file as documented in symfony2.


stof:
This should be closed, as the support for custom resource paths has been added in #166 (and in a much better way than trying to read one of the optional config files, which may not be the one holding the config you need).

and btw, this PR has nothing to do with #151 (which is talking about a custom translator class, not about a custom translation resource path)

PR #166, opened 2016-06-14
Use framework bundle translation files paths

### LINKED ISSUE #160: Read the translations folder from framework.translator.paths configuration
BazingJsTranslationBundle doesn't look into the **framework.translator.paths** for extra-paths to look for translations.

For example, I want to have my translations in infrastructure/Ui/Translations and my symfony configuration is:

```
framework:
    translator:
        fallback: en
        paths:
            - '%kernel.root_dir%/../src/Infrastructure/Ui/Translations'

```

The file **TranslationFinder** is responsible for looking the translation folder. I have hardcoded the new folder to test and it works:
```
$locations[] = $this->kernel->getRootDir() . '/../src/Infrastructure/Ui/Translations';
```

monteiro:
I noticed that in the **TranslationFinder** we were fetching by folders all the translations in the system. 
Since the FrameworkBundle already knows which translation is available I thought that we could use it. 

I've created a compiler pass to replace the argument in the **TranslationFinder** with the list of files generated by the **FrameworkBundleExtension**

I have also included a functional test to test if this works.

Closes #160 

Plz review so I can squash later the commits ;)

### DIFF
--- a/DependencyInjection/Compiler/TranslationResourceFilesPass.php
+++ b/DependencyInjection/Compiler/TranslationResourceFilesPass.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace Bazinga\Bundle\JsTranslationBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Kernel;
+
+/**
+ * @author Hugo MONTEIRO <hugo.monteiro@gmail.com>
+ */
+class TranslationResourceFilesPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('translator.default')) {
+            return;
+        }
+
+        if (Kernel::VERSION_ID < 20700) {
+            $translationFiles = $this->getTranslationFilesFromOlderSymfonyVersions($container);
+        } else {
+            $translationFiles = $this->getTranslationFiles($container);
+        }
+
+        $container->getDefinition('bazinga.jstranslation.translation_finder')->replaceArgument(0, $translationFiles);
+    }
+
+    private function getTranslationFilesFromOlderSymfonyVersions(ContainerBuilder $container)
+    {
+        $translationFiles = array();
+
+        $methodCalls = $container->getDefinition('translator.default')->getMethodCalls();
+        foreach($methodCalls as $methodCall) {
+            if ($methodCall[0] === 'addResource') {
+                $locale = $methodCall[1][2];
+                $filename = $methodCall[1][1];
+
+                if (!isset($translationFiles[$locale])) {
+                    $translationFiles[$locale] = array();
+                }
+
+                $translationFiles[$locale][] = $filename;
+            }
+        }
+
+        return $translationFiles;
+    }
+
+    private function getTranslationFiles(ContainerBuilder $container)
+    {
+        $translationFiles = array();
+
+        $translatorOptions = $container->getDefinition('translator.default')->getArgument(3);
+        if (isset($translatorOptions['resource_files'])) {
+            $translationFiles = $translatorOptions['resource_files'];
+        }
+
+        return $translationFiles;
+    }
+}
--- a/Finder/TranslationFinder.php
+++ b/Finder/TranslationFinder.php
@@ -2,26 +2,26 @@
 
 namespace Bazinga\Bundle\JsTranslationBundle\Finder;
 
-use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\Finder\Finder;
 
 /**
  * @author William DURAND <william.durand1@gmail.com>
  * @author Markus Poerschke <markus@eluceo.de>
+ * @author Hugo MONTEIRO <hugo.monteiro@gmail.com>
  */
 class TranslationFinder
 {
     /**
-     * @var KernelInterface
+     * @var array list of translation files from the framework bundle
      */
-    protected $kernel;
+    private $translationFilesByLocale;
 
     /**
-     * @param KernelInterface $kernel The kernel.
+     * @param array $translationFilesByLocale all the translations whose index is the locale
      */
-    public function __construct(KernelInterface $kernel)
+    public function __construct(array $translationFilesByLocale)
     {
-        $this->kernel = $kernel;
+        $this->translationFilesByLocale = $translationFilesByLocale;
     }
 
     /**
@@ -35,13 +35,9 @@ class TranslationFinder
      */
     public function get($domain, $locale)
     {
-        $finder = new Finder();
+        $filteredFilenames = $this->getTranslationFilesFromConfiguration($domain, $locale);
 
-        return $finder
-            ->files()
-            ->name($domain . '.' . $locale . '.*')
-            ->followLinks()
-            ->in($this->getLocations());
+        return $filteredFilenames;
     }
 
     /**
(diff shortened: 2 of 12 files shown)


Finder/TranslationFinder.php
```diff
@@ -54,59 +56,52 @@ public function all()
...
-                $locations[] = $dir;
+    /**
+     * Get all folders where there are translation files.
+     *
+     * @return array folders with translations
+     */
+    private function getFoldersWithTranslationFiles()
+    {
+        $dirToSearch = [];
+        foreach ($this->translationFilesByLocale as $locale => $resourceFilePaths) {
```
monteiro: The translationFilesByLocale comes from the FrameworkBundle and it is organized by locale => list of files in an array in the specific locale

DependencyInjection/Compiler/TranslationResourceFilesPass.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+class TranslationResourceFilesPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('translator.default')) {
+            return;
+        }
+
+        $translationFiles = [];
```
stof: The bundle currently does not require PHP 5.4 AFAIK

Finder/TranslationFinder.php
```diff
@@ -14,14 +14,14 @@ class TranslationFinder
     /**
      * @var KernelInterface
      */
-    protected $kernel;
+    protected $translationFilesByLocale;
```
stof: I suggest making the new property private, to avoid having it covered by any backward compatibility concern

DependencyInjection/Compiler/TranslationResourceFilesPass.php
monteiro: well noticed ;)

Finder/TranslationFinder.php
```diff
@@ -35,13 +35,15 @@ public function __construct(KernelInterface $kernel)
      */
     public function get($domain, $locale)
     {
-        $finder = new Finder();
+        $dirToSearch = $this->getFoldersWithTranslationFiles();
```
stof: Why are you extracting a list of folders from the list of files and then using filesystem iteration to find files again ? Just filter the array of files directly.

monteiro:
@stof I have fixed every comment you said. I have a little issue with one thing. The framework.translator.paths was introduced in Symfony 2.8.

Is there any way to workaround this situation, like deprecating old versions, or maybe using the nasty way of using the old translation finder if the symfony version is older than 2.8.

stof:
> The framework.translator.paths was introduced in Symfony 2.8.

I don't understand your issue. If you want to have a test ensuring that files configured through it are found by the bundle, you should just separate it into a dedicated test which would get skipped for older Symfony versions. Your code does not rely on this parameter

monteiro:
I did a workaround to support symfony 2.7.

Should I try to support other versions as well? I am asking this because according to the symfony version lifecycle only symfony 2.7 and up is supported.

willdurand:
Ideally, your PR should not introduce BC breaks (even if Symfony LTS is 2.7).

monteiro:
@willdurand ok, thanks. I will fix it.

monteiro:
@willdurand @stof I have fixed everything, but travis is failing on PHP 5.3 because of SSL. Is there anything we can do? 

screen shot 2016-06-15 at 15 55 18

stof:
@monteiro the 5.3.3 image on Travis is just unusable (no openssl makes things quite hard), so I think @willdurand should just remove it from the matrix (keeping the 5.3 job for the latest 5.3 version)

stof:
@monteiro what is the issue with Symfony 2.7 ? Even if there is no ``framework.translator.paths`` config settings there, it does not change the fact that the Symfony translator receives a list of resource files.
The only new feature in 2.8 is the possibility to register custom paths instead of being limited to bundle ones

monteiro:
@stof what do you mean with Symfony 2.7? 

I have put the translator: { enabled: true } in older versions, so we are sure that it gets loaded by the framework bundle.

If for some reason the list of translation files is empty, use the old way of fetching the files.

monteiro:
@stof @willdurand is there anything I could do to improve this PR, so it gets merged? :)

DependencyInjection/Compiler/TranslationResourceFilesPass.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Bazinga\Bundle\JsTranslationBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @author William DURAND <william.durand1@gmail.com>
```
willdurand: It should be you @monteiro

Finder/TranslationFinder.php
```diff
@@ -14,13 +14,19 @@ class TranslationFinder
     /**
      * @var KernelInterface
      */
-    protected $kernel;
+    private $kernel;
```
willdurand: you should probably keep `protected` here, some people use to subclass this class.

willdurand:
@monteiro I have a few remarks. Once done, you can squash your commits, and it should be fine :) Thanks!

monteiro:
@willdurand last changes done ;) Squashed all the commits and rebased. After travis runs and succeeds, it is ready to be merged!

stof:
IMO, you should remove the logic replicating the FrameworkBundle logic with your finder entirely.
The list of files is filled in the translator.default service in all versions of Symfony, even if it changed between 2.7 and 2.8. This would allow to have a single simple logic in the TranslationFinder class instead of having 2 different logic.

For Symfony 2.8+, resources are filled through the options arguments (in the ``resource_files`` option), which is what your compiler pass supports today.
For Symfony 2.7 and older, resources are added in the translator by calling ``addResource`` for each of them. The compiler pass could extract such calls from the Definition object (btw, it should do it for *all* Symfony versions, because the method is not gone even though FrameworkBundle stopped using it for resources it is adding to improve performance, because other bundles may still add such calls)

monteiro:
@stof you have a point 👍

monteiro:
@willdurand following @stof idea, I have pushed the last changes.

Now, the translation finder doesn't need the Kernel or the Finder. I'm fetching all the filenames from the translation.default definition (new and older versions of symfony).

Take a look and merge if ok ;) 
Thanks!

monteiro:
@willdurand @stof let me know if there is anything missing ;)

willdurand:
Looks good  to me

monteiro:
@stof , ready to go :+1 ?

willdurand:
thank you @monteiro

DependencyInjection/Compiler/TranslationResourceFilesPass.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private function getTranslationFiles(ContainerBuilder $container)
+    {
+        $translationFiles = array();
+
+        $translatorOptions = $container->getDefinition('translator.default')->getArgument(3);
+        if (isset($translatorOptions['resource_files'])) {
+            $translationFiles = $translatorOptions['resource_files'];
+        }
```
stof: you also need to read calls to ``addResource`` here, as other bundles can still have compiler passes adding resources through method calls

Tests/Controller/ControllerTest.php
```diff
@@ -281,4 +282,26 @@ public function testGetTranslationsWithUnderscoredLocale()
 JSON
         , $response->getContent());
     }
+
+    public function testGetTranslationsOutsideResourcesFolder()
+    {
+        if (Kernel::VERSION_ID < 20800) {
+            return;
```
stof: test should be marked as skipped instead of just returning

monteiro:
@stof thanks! 

I will create a PR to address those ;)

PR #169, opened 2016-07-07
Minor fixes of PR-166

monteiro:
Related #166 

Get translation files from add resource calls for any version of Symfony.
Use mark as skipped inside the unit test instead of a return.

//cc @stof


willdurand:
:+1:

DependencyInjection/Compiler/TranslationResourceFilesPass.php
```diff
@@ -17,16 +17,16 @@ public function process(ContainerBuilder $container)
             return;
         }
 
-        if (Kernel::VERSION_ID < 20700) {
-            $translationFiles = $this->getTranslationFilesFromOlderSymfonyVersions($container);
-        } else {
-            $translationFiles = $this->getTranslationFiles($container);
+        $translationFiles = $this->getTranslationFilesFromAddResourceCalls($container);
+
+        if (Kernel::VERSION_ID >= 20700) {
```
stof: I even suggest calling this method all the time:

- it is possible to have different versions of FrameworkBundle, HttpKernel and Translator installed, and ``Kernel::VERSION_ID`` detects the HttpKernel version here, while we deal with FrameworkBundle and Translator
- the method already returns an empty array in case the option is not defined, which will be compatible with older versions of Symfony

Tests/Controller/ControllerTest.php
```diff
@@ -286,7 +286,7 @@ public function testGetTranslationsWithUnderscoredLocale()
     public function testGetTranslationsOutsideResourcesFolder()
     {
         if (Kernel::VERSION_ID < 20800) {
-            return;
+            $this->markTestSkipped('Version do not support framework.paths configuration');
```
stof: I would say ``This Symfony version``, to explain which version you are talking about

stof:
I suggest adding a new test where a compiler pass would add its own call to ``addResource``, to avoid regressions on this

stof:
@monteiro what about the test I suggested adding ?

monteiro:
@stof It is the only thing missing in this PR. I will create it.

monteiro:
@stof I was thinking about the test. The best way to test this is to create another bundle and a compiler pass that adds the resource calls to the **translator.default** definition? 

Or there is any easy way to do it?

monteiro:
@stof I've created a bundle and inside that bundle I've created a compiler pass to simulate an addMethodCall of a definition. 

Then i've created a functional test that tries to see if that specific domain from that bundle was loaded. Since the domain is unique, everything should be fine. 

What do you think?

monteiro:
Of course I need to fix the failing tests but it works on my local :) ready for review.

monteiro:
@stof everything ready for you to review, now with tests passed ;)

monteiro:
@stof Going to merge it, what do you think? ;)

stof:
:+1:

monteiro:
Thanks again @stof !

PR #172, opened 2016-07-20
Add options to dump only specific formats and to merge all domains

bensampaio:
Adds the following command options:
* `--formats`: if set, only the passed formats will be generated;
* `--merge-domains`: if set, all domains will be merged into a single file per language.

Some examples on how the command can be used:

* `php app/console bazinga:js-translation:dump`: will have the same output as before;
* `php app/console bazinga:js-translation:dump --merge-domains`: will generate the config files and a folder for every locale. For every locale folder there will be a JS and JSON file with all domains merged;
* `php app/console bazinga:js-translation:dump --formats=js`: will generate only the JS files for every domain;
* `php app/console bazinga:js-translation:dump --formats=json --merge-domains`: will generate only the merged JSON files for every locale;


Command/DumpCommand.php
```diff
@@ -29,7 +29,27 @@ protected function configure()
...
+            ->setDescription('Dumps all JS translation files to the filesystem')
+            ->addOption(
+                'locales',
+                null,
+                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
+                'If set, only the passed locales will be generated',
+                array()
+            )
+            ->addOption(
+                'formats',
```
stof: if you use `` InputOption::VALUE_IS_ARRAY``, you should use a singular name, as the usage for multiple formats will be ``--format=js --format=json``

stof:
for the locale filtering, there is already a feature configuring the active locales in the config, and this is respected by the dump command. Do we actually need a second filtering layer at dump-time ?

bensampaio:
you are right, I didn't notice that in the documentation when I looked for it but I saw it now... sorry for that.

In my last commit I removed the `locales` option, changed the `formats` option to `format`, and changed the tests. Thanks!

monteiro:
@bensampaio can you also put this new feature explained in Resources/doc/index.md ? 

Thanks!

bensampaio:
@monteiro done :)

Dumper/TranslationDumper.php
```diff
@@ -118,21 +118,38 @@ public function addLoader($id, $loader)
      * Dump all translation files.
      *
      * @param string $target Target directory.
+     * @param array $formats Formats to generate.
```
stof: I would document it as ``string[]`` to give more details

monteiro:
Nice! If @stof agrees, since he also commented the PR, I can merge ;)

Dumper/TranslationDumper.php
```diff
@@ -194,6 +211,41 @@ private function dumpTranslations($route, array $formats, $target)
...
+                    )
+                );
+
+                $file = sprintf(
+                    '%s/%s',
+                    $target,
+                    strtr(
+                        $route->getPath(),
+                        array(
+                            '{domain}' => sprintf('%s/%s', $locale, $locale),
```
stof: why using the locale twice as domain ?

Command/DumpCommand.php
```diff
@@ -29,7 +29,20 @@ protected function configure()
...
+            ->setDescription('Dumps all JS translation files to the filesystem')
+            ->addOption(
+                'format',
+                null,
+                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
+                'If set, only the passed formats will be generated',
+                array()
+            )
+            ->addOption(
+                'merge',
```
stof: this should maybe be named ``merge-domains``, as you may as well merge locales in the future

Dumper/TranslationDumper.php
bensampaio: because you can have two files for the same locale: `en.js` and `en.json`. I think it looks better if they are under a folder.

stof: well, I don't really agree here (and some people might prefer grouping all JS files than grouping all english files btw)

bensampaio: So what do you propose? All JS and JSON files under `translations` folder?

I agree that some people might prefer grouping the files by format, but that is true not just for this case. When dumping translations per domain you could also prefer that instead of the current solution (a folder per domain). For that reason, I think that option shouldn't be considered now.

bensampaio: @stof please let me know your thoughts on this. Should I leave it as it is or not create the locale folder? Thanks!

stof: I vote for not creating the folder.

But I'm not the maintainer of this bundle, so @monteiro should have the last word on it.

monteiro: @bensampaio creating the folder is good. Sorry for the delay!

bensampaio: Ok, thanks @monteiro. Can this be merged then? Or is there anything missing?

monteiro: Sorry @bensampaio. What I wanted to say was that we should remove the duplication of folders like @stof was saying.

I don't see a clear advantage on using folders here, specially duplicating the name of the file. We could use folders if the name didn't have the locale. But in this case it has, so we can remove that duplication.

Sorry for my mistake.

What do you think?

bensampaio: Ok, I personally prefer the folder but I also agree it looks weird to have a folder and files with the same name. So I'll remove it :)

bensampaio:
I removed the locale folder as decided. Merged translations are now placed under the `translations` folder.

bensampaio:
I realised there was a bug in the way the `$merge` flag was being accessed, but I fixed it in the last commit.

monteiro:
@bensampaio so good to go? //cc @stof

bensampaio:
@monteiro I realised today that the `TranslationFinder` in `master` is not working for me. The tests run but when I change my composer config to load `dev-master` no translation files are found after running the dump command. Is this a bug?

I tried to merge my changes with `v2.5.0` and everything seems to work as expected.

monteiro:
@bensampaio I need to test it. It is true that the last PR that I did is related with that. 

Which version of symfony are you using where this fails?

monteiro:
And where are the translation files so I can replicate it in my environment.

bensampaio:
@monteiro my version of Symfony is 2.8.9 and my folder structure is the following:

`src/<SomeName>/<BundleName>/Resources/translations/<domain>.<locale>.xliff`

Let me know if I can help with more info. Hope it's easy to fix!

monteiro:
@bensampaio tomorrow morning I am going to test it. I will let you know.

monteiro:
@bensampaio I've tested with Symfony 2.8.3 and 2.8.10, everything worked perfectly.

In my company, we also use the same directory structure as you have inside **AppBundle/Resources/translations**

If you can please debug the **TranslationFinder**. The constructor receives all the folders with the translations and that parameter is generated in the dependency injection step, so this is cached. Remove cache, and tell me if it persists.

monteiro:
@bensampaio can you share somehow the composer.json and a little example of the translation files you have? If you still have problems.

bensampaio:
@monteiro I cleared the cache but the problem persists. I debugged the `TranslationFinder` and it seems to always receive an empty list.

I cannot share the `composer.json`. Is there anything important in particular you need? Some information that might be helpful:

- In the `require` list we have: `"willdurand/js-translation-bundle": "dev-master"` (I tried both with your repo and my forked repo and both have the same problem).
- I don't know if this is relevant but in our `autoload` section we have:

```
"psr-4": {
    "ProjectName\\": "src/SomeName/"
}
```

As you requested below there is an example of a translation file:

```
<?xml version="1.0" encoding="utf-8"?>
<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
    <file source-language="en" target-language="en" datatype="plaintext" original="not.available">
        <body>
            <trans-unit id="some_id" resname="title">
                <source>title</source>
                <target>Hello World!</target>
            </trans-unit>
        </body>
    </file>
</xliff>
```

Let me know if this helps or if you need any more info. Thanks!

bensampaio:
I found the source of the problem. The translator was disabled in the configuration for some reason unknown to me. Sorry for the confusion.

Well problem fixed so I think you can proceed with the merge, right? Thanks!

bensampaio:
Awesome, thanks! Any plans to create a new release with the last changes?

monteiro:
@bensampaio sure! in the next days ;)

PR #173, opened 2016-07-20
Add npm configuration for publishing

bensampaio:
Added the file `Resources/package.json`.

If you run `npm publish` in the `Resources` folder, the package will be published only with the folders `js` and `public/js`. This way the JS is still part of the bundle but is also available through npm.

Making the 'Translator' available through npm allows JS developers to `require` / `import` it when using `webpack`, `jspm`, or other modules bundler. And also allows the separation of back-end and front-end dependencies.


monteiro:
@bensampaio yes, you are right. I am not using webpack to import this module, but I should as well. 

Can you write a little section in https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/doc/index.md to explain how can we import this via webpack for example?

Resources/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "bazinga-translator",
+  "version": "1.0.0",
```
monteiro: The version used should be the same as the BazingaJsBundle. That way we keep both on track, if in the future we want to update the javascript. Like reactjs does in all of their internal components.

Resources/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "email": "will+git@drnd.me"
+  },
+  "contributors": [
+    {
+      "name": "Bruno Sampaio",
+      "email": "bens.sampaio@gmail.com"
+    }
+  ],
+  "files": [
+    "js",
```
stof: you should include only ``js/translator.js``. The folder includes a bunch of other files which are needed only by the testsuite

bensampaio:
I changed the version in the `package.json` and included only the `translator` files in it.

monteiro:
@bensampaio now the only thing missing is to explain it in https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/doc/index.md :)

bensampaio:
@monteiro I just did that, what do you think?

Resources/doc/index.md
```diff
@@ -44,16 +46,36 @@ Publish assets:
 
     php app/console assets:install --symlink web
 
+### Require via NPM
```
stof: this is missing an explanation saying that it is optional (it is only needed when wanting to import the translator using modules).

and you should also document that it is very strongly recommended to keep the version in sync between the npm package and the bundle version being used.

bensampaio: Just did that ;) thanks!

monteiro:
@bensampaio Awesome! Thanks!

bensampaio:
Thanks! Don't forget to run `npm publish` on the `Resources` folder so it becomes available on npm :)

monteiro:
@bensampaio Thanks! I will do it tomorrow ;)

monteiro:
@bensampaio already published

screen shot 2016-07-26 at 13 47 54

PR #179, opened 2016-09-14
Improve the JS so it supports NodeJS

bensampaio:
I came across this problem when trying to implement server rendering on a ReactJS application dependent on the bazinga-translator. Since the variables `window` and `document` don't exist in NodeJS, the Translator initialization would always fail.

To fix that, I improved the module system detection so that when required on NodeJS the code no longer fails because `document` and `window` don't exist. Also removed the dependency on the variable `undefined`.

Besides that, I added `uglify-js` as a dependency and the build script so other contributors don't need to guess which uglification options should be used. To uglify the JS now you just need to run `npm run build` under the `Resources` directory.

What do you think?


bensampaio:
I forgot to mention that everything should work as before, the code was just rearranged.

monteiro:
@bensampaio it would be good for you to explain the problem you are having at your work as an example in the documentation. 

The PR seems OK. I'm going to merge it.

Resources/js/translator.js
```diff
@@ -401,202 +597,10 @@ var Translator = (function(document, undefined) {
...
-        /**
-         * @api public
-         */
-        reset: function() {
-            _messages   = {};
-            _domains    = [];
-            this.locale = get_current_locale();
+        if (typeof document !== 'undefined') {
+            return document.documentElement.lang.replace('-', '_');
         }
```
stof: you should return ``en`` as default locale (or use a different detection strategy). Putting it as ``null`` is a bad idea.

Resources/public/js/translator.min.js
```diff
@@ -1,5 +1 @@
-/*!
- * William DURAND <william.durand1@gmail.com>
- * MIT Licensed
- */
```
stof: please update the build script so that it keeps the license header

bensampaio:
@stof, sorry for the delay on an answer. Since this was already merged, how do you propose I make the changes? I no longer have the forked repo.

stof:
@bensampaio create a new PR

pierredup:
This breaks when using RequireJs.
The dumped translations expects the global window variable `Translator` to be defined, which won't be the case anymore with this change when loading the translation.min.js through RequireJs

PR #182, opened 2016-10-05
Remove short array sintax to support php53

### LINKED ISSUE #181: V2.6.0 and PHP 5.3
Hello,
Since v2.6.0, this bundle in not compatible with php 5.3 anymore.
No more Travis  on php 5.3
And short array syntax introduced in code base in PR #172 

The issue is that composer.json has no php constraint and old project using this bundle with constraint like "^2.0" will update to 2.6.0

monteiro:
closes #181 

//cc @JakeFr

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,8 @@ matrix:
       - php: 5.6
         env: SYMFONY_VERSION='~2.3.0'
       - php: 5.6
-        env: SYMFONY_VERSION='~2.5.0'
-      - php: 5.6
         env: SYMFONY_VERSION='~2.7.0'
       - php: 5.6
         env: SYMFONY_VERSION='~2.8.0'
       - php: 5.6
-        env: SYMFONY_VERSION='~3.0.0'
+        env: SYMFONY_VERSION='~3.1.5'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: php
 
 php:
+    - 5.3
+    - 5.4
     - 5.5
     - 5.6
     - 7.0
@@ -17,7 +19,7 @@ before_install:
     - if [ "$SYMFONY_VERSION" != "" ]; then composer require symfony/symfony:${SYMFONY_VERSION}; fi;
 
 install:
-    - composer install
+    - composer install --prefer-source
 
 script:
     - phpunit --coverage-text
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
     - if [ "$SYMFONY_VERSION" != "" ]; then composer require symfony/symfony:${SYMFONY_VERSION}; fi;
 
 install:
-    - composer install --prefer-source
+    - composer install
 
 script:
     - phpunit --coverage-text
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,6 @@ matrix:
     fast_finish: true
     include:
       - php: 5.6
-        env: SYMFONY_VERSION='~2.3.0'
-      - php: 5.6
         env: SYMFONY_VERSION='~2.7.0'
       - php: 5.6
         env: SYMFONY_VERSION='~2.8.0'
-      - php: 5.6
-        env: SYMFONY_VERSION='~3.1.5'
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -24,7 +24,4 @@
             </exclude>
         </whitelist>
     </filter>
-    <php>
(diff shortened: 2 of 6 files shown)


stof:
@monteiro I suggest adding the PHP requirement too, as it serves as a documentation purpose about supported versions too (and it avoids forgetting to add it when dropping support for older versions)

JakeFr:
I found some other short array syntax
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Dumper/TranslationDumper.php#L124
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Tests/Dumper/TranslationDumperTest.php#L217
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Tests/Dumper/TranslationDumperTest.php#L251
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Tests/Dumper/TranslationDumperTest.php#L293
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Tests/Dumper/TranslationDumperTest.php#L335
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Tests/Dumper/TranslationDumperTest.php#L363

.travis.yml
```diff
@@ -17,7 +19,7 @@ before_install:
     - if [ "$SYMFONY_VERSION" != "" ]; then composer require symfony/symfony:${SYMFONY_VERSION}; fi;
 
 install:
-    - composer install
+    - composer install --prefer-source
```
monteiro: https://twitter.com/travisci/status/403240648842964992

phpunit.xml.dist
```diff
@@ -24,4 +24,7 @@
             </exclude>
         </whitelist>
     </filter>
+    <php>
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
```
monteiro: https://github.com/symfony/symfony/issues/18222

monteiro:
@stof I did the changes to travis.yml with support for 5.3 and 5.4. 

Also I have added the necessary changes because of the deprecation notices breaking the travis build because of phpunit bridge deprecation handler.

.travis.yml
monteiro: I have already fixed this.

phpunit.xml.dist
stof: it should not always be weak. We precisely want to prevent using deprecated APIs (users will complain if you use such API).
You should fix the usage of deprecated APIs instead

monteiro: @stof What do you mean by fix usage of deprecated APIs?

Travis build breaks only in Symfony 3.0. We can still show the deprecation warnings, but avoiding it breaking the travis build.

monteiro: Maybe forcing Symfony 3.0 to use the latest Twig in this case avoiding deprecation warnings?

stof: and if you insist on testing on unmaintained Symfony version, Travis should set the env variable only for these jobs in the matrix, not for all jobs (so that tests break if the bundle is using a deprecated API)

.travis.yml
```diff
@@ -32,10 +32,8 @@ matrix:
       - php: 5.6
         env: SYMFONY_VERSION='~2.3.0'
```
stof: 2.3 is not maintained anymore either

.travis.yml
```diff
@@ -32,10 +32,8 @@ matrix:
...
       - php: 5.6
-        env: SYMFONY_VERSION='~2.5.0'
-      - php: 5.6
         env: SYMFONY_VERSION='~2.7.0'
       - php: 5.6
         env: SYMFONY_VERSION='~2.8.0'
       - php: 5.6
-        env: SYMFONY_VERSION='~3.0.0'
+        env: SYMFONY_VERSION='~3.1.5'
```
stof: I would not include 3.1 at all. It is the latest stable, so it is covered by the normal tests already, and it will become irrelevant very fast after normal tests are bumping to 3.2 when it gets released. My own policy is to add only LTS version as extra jobs.

.travis.yml
monteiro: True

.travis.yml
monteiro: @stof should I include the latest of 3.0, or 2.8 it is enough?

monteiro: Nevermind. I understand what you have said.

monteiro:
@stof everything is fixed and working. Ready to be merged, what do you think?

PR #183, opened 2016-10-13
Fix for some problems introduced by PR #179

bensampaio:
As suggested by @stof:

- `get_current_locale` now returns the fallback locale if none is found.
- The build script now keeps the license code block in the top of the file.


stof:
:+1:

PR #185, opened 2016-11-18
Add default count parameter on transchoice

pyrech:
The idea is to have a similar behavior between Symfony translator ([TranslationExtension](https://github.com/symfony/symfony/blob/5129c4cf7e294b1a5ea30d6fec6e89b75396dcd2/src/Symfony/Bridge/Twig/Extension/TranslationExtension.php#L98) and https://github.com/symfony/symfony/pull/19795) and the js translator.

### DIFF
--- a/Resources/js/translator.js
+++ b/Resources/js/translator.js
@@ -149,6 +149,11 @@
             );
 
             var _number  = parseInt(number, 10);
+            parameters = parameters || {};
+
+            if (parameters.count === undefined) {
+                parameters.count = number;
+            }
 
             if (typeof _message !== 'undefined' && !isNaN(_number)) {
                 _message = pluralize(
@@ -158,7 +163,7 @@
                 );
             }
 
-            return replace_placeholders(_message, parameters || {});
+            return replace_placeholders(_message, parameters);
         },
 
         /**
--- a/Resources/js/translatorTest.js
+++ b/Resources/js/translatorTest.js
@@ -52,7 +52,7 @@ test('trans()', function() {
 });
 
 test('transChoice()', function() {
-    expect(26);
+    expect(31);
 
     Translator.add('foo.plural', '{0} Nothing|[1,Inf[ Many things', 'Foo');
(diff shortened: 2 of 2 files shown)


pyrech:
Hi @monteiro and @willdurand :) Will you have a look at this PR soon?

Resources/js/translator.js
```diff
@@ -149,6 +149,11 @@
             );
 
             var _number  = parseInt(number, 10);
+            parameters = parameters || {};
+
+            if (!parameters.count) {
```
stof: this is wrong. You need to check whether it is defined or no. ``0`` should not be replaced if it is passed explicitly.

stof: Please also add tests covering the case where the count is provided explicitly too

pyrech: Right, fixed now

pyrech:
Hi

willdurand:
looks good to me!

monteiro:
@pyrech thanks for the pr! The tests seems to be failing :( can you fix it? So we can merge this.

pyrech:
Tests were working before I rebased the PR. The error is `Asset support cannot be enabled as the Asset component is not installed`. The main difference I see is that Circle now installs Symfony 3.2 so I guess it looks like the test suite is not compatible with the latest release of Symfony

monteiro:
I see what you mean @pyrech

monteiro:
@pyrech we are already fixing the tests problem: https://github.com/willdurand/BazingaJsTranslationBundle/pull/186 After this is merged we make this tests green again.

monteiro:
@pyrech Can you rebase with master, so we can merge this PR with tests passed? Thanks a lot!

pyrech:
Cool :+1: Just rebased the PR on top of master :wink:

monteiro:
Thanks a lot!

pyrech:
Awesome, you're welcome :tada:

PR #186, opened 2016-12-05
Add missing dependencies to composer.json (because Symfony 3.2 framework bundle removed them)

monteiro:
The new Symfony 3.2 removed some dependencies from framework bundle:
Related: https://github.com/symfony/symfony/pull/20067

Also added phpunit to composer.json because of: https://github.com/sebastianbergmann/phpunit/pull/2264

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
     - composer install
 
 script:
-    - phpunit --coverage-text
+    - vendor/bin/phpunit --coverage-text
     - phantomjs Resources/js/run-qunit.js file://`pwd`/Resources/js/index.html
     - phantomjs Resources/js/run-qunit.js file://`pwd`/Resources/js/index-with-es5-shim.html
 
--- a/Tests/Fixtures/app/config/older_versions_config.yml
+++ b/Tests/Fixtures/app/config/older_versions_config.yml
@@ -4,6 +4,7 @@ framework:
     templating:    { engines: [ 'twig' ] }
     test:          ~
     translator:    { enabled: true }
+    validation:    { enabled: false }
 
 # Twig Configuration
 twig:
--- a/composer.json
+++ b/composer.json
@@ -14,13 +14,17 @@
         "symfony/framework-bundle": "~2.7|~3.1",
         "symfony/finder": "~2.7|~3.1",
         "symfony/console": "~2.7|~3.1",
-        "symfony/intl": "~2.7|~3.1"
+        "symfony/intl": "~2.7|~3.1",
+        "symfony/templating": "~2.7|~3.1",
+        "symfony/translation":  "~2.7|~3.1"
     },
     "require-dev": {
+        "symfony/asset": "~2.7|~3.1",
         "symfony/yaml": "~2.7|~3.1",
         "symfony/browser-kit": "~2.7|~3.1",
         "symfony/twig-bundle": "~2.7|~3.1",
-        "symfony/phpunit-bridge": "~2.7|~3.1"
+        "symfony/phpunit-bridge": "~2.7|~3.1",
+        "phpunit/phpunit": "^4.8|~5.7"
     },
     "replace": {
         "willdurand/expose-translation-bundle": "2.5.*"
--- a/Tests/Fixtures/app/AppKernel.php
+++ b/Tests/Fixtures/app/AppKernel.php
@@ -66,6 +66,10 @@ class AppKernel extends Kernel
         } else {
             $loader->load(__DIR__.'/config/'.$this->environment.'.yml');
         }
+
+        if (self::VERSION_ID > 30200) {
+            $loader->load(__DIR__.'/config/disable_annotations.yml');
+        }
     }
 
     public function serialize()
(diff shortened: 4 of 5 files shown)


stof:
your PR is wrong. It does not add support for 3.2. It drops support for 3.1

monteiro:
@stof so in this case, what can I do to solve this situation?

stof:
don't change ``~3.1`` to ``~3.2``. The ``~3.1`` constraint already allows 3.2 (as well as any newer 3.x version)

monteiro:
@stof what about requiring explicitly symfony/asset?

monteiro:
@stof can you take a look for a possible solution in this PR? All of my decisions were based in the urls in the PR's description.

composer.json
```diff
@@ -14,13 +14,18 @@
         "symfony/framework-bundle": "~2.7|~3.1",
         "symfony/finder": "~2.7|~3.1",
         "symfony/console": "~2.7|~3.1",
-        "symfony/intl": "~2.7|~3.1"
+        "symfony/intl": "~2.7|~3.1",
+        "symfony/asset": "~2.7|~3.1",
+        "symfony/templating": "~2.7|~3.1",
+        "symfony/translation":  "~2.7|~3.1",
+        "doctrine/annotations": "~1.0"
```
stof: This bundle does not rely on annotations, so there is no reason to make them mandatory requirements

composer.json
```diff
@@ -14,13 +14,18 @@
         "symfony/framework-bundle": "~2.7|~3.1",
         "symfony/finder": "~2.7|~3.1",
         "symfony/console": "~2.7|~3.1",
-        "symfony/intl": "~2.7|~3.1"
+        "symfony/intl": "~2.7|~3.1",
+        "symfony/asset": "~2.7|~3.1",
```
stof: this should not be a mandatory requirement (it may be needed as a dev requirement if the testsuite uses it).
The bundle does not forces you to use the Asset component to load the JS file.

composer.json
```diff
@@ -14,13 +14,18 @@
...
+        "symfony/translation":  "~2.7|~3.1",
+        "doctrine/annotations": "~1.0"
     },
     "require-dev": {
         "symfony/yaml": "~2.7|~3.1",
         "symfony/browser-kit": "~2.7|~3.1",
         "symfony/twig-bundle": "~2.7|~3.1",
-        "symfony/phpunit-bridge": "~2.7|~3.1"
+        "symfony/phpunit-bridge": "~2.7|~3.1",
+        "phpunit/phpunit": "^4.8"
```
stof: allowing only 4.8 is a bad idea, as PHPunit 4.8 does not officially support PHP 7 (and will break totally on PHP 7.1 when mocking some classes). Maintained versions of PHPUnit (i.e. 5.x) must be allowed by the constraint.

monteiro: To support 5.x version we need to drop support for PHP 5.5.

```
  Problem 1
    - phpunit/phpunit 5.7.2 requires php ^5.6 || ^7.0 -> your PHP version (5.5.21) does not satisfy that requirement.
    - phpunit/phpunit 5.7.1 requires php ^5.6 || ^7.0 -> your PHP version (5.5.21) does not satisfy that requirement.
    - phpunit/phpunit 5.7.0 requires php ^5.6 || ^7.0 -> your PHP version (5.5.21) does not satisfy that requirement.
    - Installation request for phpunit/phpunit ~5.7 -> satisfiable by phpunit/phpunit[5.7.0, 5.7.1, 5.7.2].

```

monteiro:
@stof can you take a look again? :)

monteiro:
@stof good to merge?

.travis.yml
```diff
@@ -22,7 +22,7 @@ install:
     - composer install
 
 script:
-    - phpunit --coverage-text
+    - vendor/phpunit/phpunit/phpunit --coverage-text
```
stof: use ``vendor/bin/phpunit`` instead of relying on the path inside phpunit itself (which is wrong btw, this is *not* the file exposed as a composer bin in all versions)

composer.json
```diff
@@ -14,13 +14,18 @@
...
+        "symfony/translation":  "~2.7|~3.1"
     },
     "require-dev": {
+        "symfony/asset": "~2.7|~3.1",
         "symfony/yaml": "~2.7|~3.1",
         "symfony/browser-kit": "~2.7|~3.1",
         "symfony/twig-bundle": "~2.7|~3.1",
-        "symfony/phpunit-bridge": "~2.7|~3.1"
+        "symfony/phpunit-bridge": "~2.7|~3.1",
+        "doctrine/annotations": "~1.0",
```
stof: do we really need annotations ?

monteiro:
@stof I tried to remove the doctrine annotations, but I get always the same error. That we miss that dependency. I tried even to disable it from the configuration of the tests. Am I missing something?

Tests/Fixtures/app/config/older_versions_config.yml
```diff
@@ -4,6 +4,8 @@ framework:
     templating:    { engines: [ 'twig' ] }
     test:          ~
     translator:    { enabled: true }
+    validation:    { enabled: false, enable_annotations: false }
```
stof: ``enable_annotations: false`` is useless if you disable it entirely

stof:
looks good to me

PR #189, opened 2016-12-22
Implements country support and language fallback (sf2 compliant)

Maxooo:
This feature implements <s>region</s> country support and language fallback as described in the Symfony2/3 documentation (http://symfony.com/doc/current/translation.html#fallback-translation-locales).

This feature refers to the https://github.com/willdurand/BazingaJsTranslationBundle/issues/135 issue/feature.


Resources/public/js/translator.min.js
```diff
@@ -2,4 +2,39 @@
...
\ No newline at end of file
+!function(e,a){"function"==typeof define&&define.amd?define("Translator",a):"object"==typeof module&&module.exports?module.exports=a():e.Translator=a()}(this,function(){"use strict"
```
bdunogier: Is this file maintained minified ? :)

monteiro:
@Maxooo hi! 

Can you add a test in TranslatorTest to handle this case? Thanks a lot!

Maxooo:
Thank you @monteiro for your answer.
JS is not my predilection coding language, but I'll do my best to cover this case with a test.

Thanks again !

Maxooo:
Hi @monteiro !

Sorry for the delay, I finally took the time to write the test case.
The new test seems quite similar to the 13rd (searches in part domain, if not exists in full domain) but it differs and check the right fallback message.

Hope you will like it ;)

monteiro:
Thanks a lot @Maxooo

Maxooo:
Woo fast merge ! Thank you !

goodhanded:
Looking at the test. This works if you have a 'de' translation, but what if you have a 'de_CH' in addition to the 'de'? It appears to only use the latter.

Maxooo:
Hi @goodhanded and sorry for my late answer. Could you be more precise on the expected behavior and the one you get instead ? Sorry I'm not sure to well understand.

PR #192, opened 2017-01-30
Remove router dependency from translation dumper.

### LINKED ISSUE #190: Dependency of TranslationDumper on BazingaJs route
When  I remove the route the dumper does not work. Maybe I just want to fetch the static translation file and not use the controller.

Goal: Investigate if we can remove this route dependency. 

```
Installing translation files in web/js directory
PHP Fatal error:  Call to a member function getRequirements() on null in /var/snap-ci/repo/sah-build/vendor/willdurand/js-translation-bundle/Bazinga/Bundle/JsTranslationBundle/Dumper/TranslationDumper.php on line 125

Fatal error: Call to a member function getRequirements() on null in /var/snap-ci/repo/sah-build/vendor/willdurand/js-translation-bundle/Bazinga/Bundle/JsTranslationBundle/Dumper/TranslationDumper.php on line 125

                                                              
  [Symfony\Component\Debug\Exception\FatalErrorException]     
  Error: Call to a member function getRequirements() on null  
                                                              

bazinga:js-translation:dump [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-s|--shell] [--process-isolation] [-e|--env ENV] [--no-debug] [--] <command> [<target>]
```

monteiro:
closes #190 

In some cases we don’t want to add the bazingaJs routes because we just generate the static file with all the translations in order to avoid the amount of requests to generate the file.

### DIFF
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -35,7 +35,7 @@ class DumpCommand extends ContainerAwareCommand
             ->addOption(
                 'pattern',
                 null,
-                InputOption::VALUE_REQUIRED,
+                InputOption::VALUE_OPTIONAL,
                 'the route pattern',
                 TranslationDumper::DEFAULT_TRANSLATION_PATTERN
             )
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -10,6 +10,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * @author Adrien Russo <adrien.russo.qc@gmail.com>
+ * @author Hugo Monteiro <hugo.monteiro@gmail.com>
  */
 class DumpCommand extends ContainerAwareCommand
 {
@@ -42,6 +43,11 @@ class DumpCommand extends ContainerAwareCommand
                 null,
                 InputOption::VALUE_NONE,
                 'If set, all domains will be merged into a single file per language'
+            )->addOption(
+                'path',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'If set, all the translations are in a different folder (default is "translations")'
             );
     }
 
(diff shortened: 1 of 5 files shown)


Command/DumpCommand.php
```diff
@@ -81,6 +87,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            ->dump($this->targetPath, $this->getTranslationsPath($input->getOption('path')), $formats, $merge);
+    }
+
+    /**
+     * Get translations path from the router defined or use the default one.
+     */
+    private function getTranslationsPath($path)
+    {
+        $translationsPathOverriden = empty($path) ? 'translations' : $path;
+        $route = $this->getContainer()->get('router')->getRouteCollection()->get('bazinga_jstranslation_js');
```
stof: this does not remove the dependency, as you always get it

monteiro: This is what normally we need to add to a routing.yml to make BazingaJs to work:
```
_bazinga_jstranslation:
    resource: "@BazingaJsTranslationBundle/Resources/config/routing/routing.yml"
```

If we remove this route from the routing.yml will this still be available in the route collection?

stof: nope, as this is what is adding it.

You should use the router only when the CLI option is not used at all, and make the option pass the full pattern you need.

monteiro: so the CLI option only receives a pattern like ```/translations/{domain}.{_format}```.

TranslationDumper receives this pattern if the CLI is used. If it does not receive it (meaning that the CLI was not used), it should use the router pattern instead. Right?

stof: the dumper should require getting the pattern as argument IMO (so no dependency on the router).

And the Command class should have an option to pass this pattern. If the option is not used, it should get the pattern from the router (i.e. optional deps on the router here)

monteiro:
@stof sorry for the time :) But I did some changes to remove the router from the CLI command according to your suggestion.

Command/DumpCommand.php
```diff
@@ -31,6 +33,13 @@ protected function configure()
             ))
             ->setDescription('Dumps all JS translation files to the filesystem')
             ->addOption(
+                'pattern',
+                null,
+                InputOption::VALUE_OPTIONAL,
```
stof: that's wrong. The value is required. Passing `--pattern` without value does not make sense.

monteiro:
@stof fixed

PR #201, opened 2017-04-11
Add support for Symfony 3.3

pierredup:
Since symfony/symfony#22010 there is now an extra argument to the translator constructor, which means the `$options` parameter moves one index up

### DIFF
--- a/DependencyInjection/Compiler/TranslationResourceFilesPass.php
+++ b/DependencyInjection/Compiler/TranslationResourceFilesPass.php
@@ -4,6 +4,7 @@ namespace Bazinga\Bundle\JsTranslationBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\OutOfBoundsException;
 
 /**
  * @author Hugo MONTEIRO <hugo.monteiro@gmail.com>
@@ -46,8 +47,14 @@ class TranslationResourceFilesPass implements CompilerPassInterface
     private function getTranslationFiles(ContainerBuilder $container)
     {
         $translationFiles = array();
+        $translator = $container->findDefinition('translator.default');
 
-        $translatorOptions = $container->findDefinition('translator.default')->getArgument(3);
+        try {
+            $translatorOptions = $translator->getArgument(4);
+        } catch (OutOfBoundsException $e) {
+            $translatorOptions = $translator->getArgument(3);
+        }
+        
         if (isset($translatorOptions['resource_files'])) {
             $translationFiles = $translatorOptions['resource_files'];
         }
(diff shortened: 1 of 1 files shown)


monteiro:
@pierredup thank you very much for your PR. Thanks to you we are supporting the next version of Symfony.

stof:
The code should not be merging both arguments, but reading the right one.

stof:
see https://github.com/Incenteev/translation-checker-bundle/pull/11 for a solution

pierredup:
It only merges the values to check for the `resource_files` option in either the 3rd or 4th argument in the compiler pass. It doesn't pass those merged options back to to the service

stof:
@pierredup yeah, but this break if the other argument is not the options argument

craigh:
see #293

pierredup:
@stof how does it break? This change covers all scenarios for SF < 3.3 as well as >= 3.3

pierredup:
Okay I just read through all the linked issues. The issue here is not with SF >= 3.3, but rather 5.2 where the arguments are not resolved yet by the time this compiler pass runs. 
And since this bundle don't support SF < 3.4 anymore, I think the compiler pass can be updated to just rely on the new signature and does'n have to worry about the old signature

craigh:
this argument:

>  The issue here is not with SF >= 3.3, but rather 5.2 where the arguments are not resolved yet by the time this compiler pass runs

is opposite this argument:

> I think the compiler pass can be updated to just rely on the new signature and does'n have to worry about the old signature


if the issue is the timing, then the order is correct. If the issue is the new signature order then timing doesn't matter.

pierredup:
In Symfony < 5.2, the 4th argument (3rd index) was an array. In Symfony 5.2, that changed to be an instance of `Symfony\Component\DependencyInjection\Argument\AbstractArgument`.

This compiler pass runs before the `Symfony\Component\Translation\DependencyInjection\TranslatorPass`, which replaces the `AbstractArgument` with the array of loader ids.



> I think the compiler pass can be updated to just rely on the new signature

What I meant is the new signature from Symfony 4, not 5.2. So this bundle can drop support for the Symfony < 3.3 signature, and just rely on the the 4th index to be the options.

> this argument:
> 
> > The issue here is not with SF >= 3.3, but rather 5.2 where the arguments are not resolved yet by the time this compiler pass runs
> 
> is opposite this argument:
> 
> > I think the compiler pass can be updated to just rely on the new signature and does'n have to worry about the old signature
> 
> if the issue is the timing, then the order is correct. If the issue is the new signature order then timing doesn't matter.

What I meant with the new signature, is the one from Symfony 3.3, not 5.2. So if this bundle wanted to keep support for Symfony < 3.3, then the compiler pass needs to be updated to run after the Symfony compiler pass, so that the arguments can be resolved before this pass runs. But this bundle dropped support for Symfony <= 3.3 in #271, so the pass needs to be updated to just check for the 4th index and doesn't have to cater for the 3rd index anymore.

PR #202, opened 2017-04-19
Remove unnecessary use of `Translator` in getTranslations.js.twig

ruudk:
No need to call `Translator` when it's injected already. Just use short `t`. 

For larger files it will help shave of some bytes on every request :)

### DIFF
--- a/Resources/views/config.js.twig
+++ b/Resources/views/config.js.twig
@@ -1,4 +1,4 @@
-(function (Translator) {
-    Translator.fallback      = '{{ fallback }}';
-    Translator.defaultDomain = '{{ defaultDomain }}';
+(function (t) {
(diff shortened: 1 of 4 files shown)


monteiro:
@ruudk In this case you're just renaming Translator for t. There is no need for that because uglify also compresses the file. Have you seen translator.min.js? You can generate it when you run "npm run build".

stof:
@monteiro this is not about the translator file itself, but about the translation files which are generated on the fly or dumped. These ones are not uglified

ruudk:
@stof Indeed. I include `bazinga_jstranslation_js` in every request. That one is automatically generated and cached.

stof:
👍 for these changes

PR #203, opened 2017-05-01
Add pattern default option in DumpCommand + Update documentation

monteiro:
- Updated documentation related with the new pattern argument
- I've added a default value to the pattern option in DumperCommand. This way it keeps being backward compatible. Thanks to @kcivey.


kcivey:
Can't it just have a default value, so it's backward-compatible (#204)?

PR #204, opened 2017-05-01
add missing default for pattern in DumpCommand

kcivey:
Without this, I get this error if I don't use the `--pattern` option (it uses an empty pattern):

> [Symfony\Component\Debug\Exception\ContextErrorException]                                         
  Warning: file_put_contents(.../web/js/): failed to open stream: Is a directory


monteiro:
@kcivey I've updated documentation and added the default option to the default option in #203

monteiro:
Your change is already in master. Going to close this PR.

PR #207, opened 2017-07-24
Fix IE bug related to $0 in target placeholder value

bOmBeLq:
Translating message `%value%/hr` with `{value: $0.01}` arguments will result in
`%value%.00/hr` because:
1. `%value%/hr` is replaced to `$0.01/hr`
2. `$0` is treated as special reg-exp char (full regexp match)
3. So `$0` is replaced with `%value%`.

Please type:
`'%value%/hr'.replace(new RegExp('%value%', 'g'), '$0.01')`
in ie11 console to get better clue.

I've added the test-case but if you would be about to remove fixes from translator.js and keep the test then it would still pass without the fix. To make it fail need to run test on IE 11 engine somehow.


bOmBeLq:
@mrtorrent @willdurand I am not sure why the automated tests have failed. It seems like my commits are not related to it:
https://travis-ci.org/willdurand/BazingaJsTranslationBundle/jobs/256948601

willdurand:
@bOmBeLq could you rebase your PR?

monteiro:
@bOmBeLq if you rebase everything should be green because of this commit: https://github.com/willdurand/BazingaJsTranslationBundle/commit/60a9f7de3a492194511fe50f95f6e71acbd28944

bOmBeLq:
Thank you. Seems like rebase helped.

bOmBeLq:
I have squashed the commits. Let me know if need to do anything else. 
Thanks for your support guys :)

PR #211, opened 2017-08-04
Move the license file at the root

stof:
Having the license at the root of the repo allows Github to detect it, displaying the license in the repo heading and exposing it in the API.
The Symfony best practice for shared bundles changed several months ago, to recommend putting the license at the root rather than in `Resources/meta` precisely to follow the common convention now supported by Github.


willdurand:
thanks!

PR #212, opened 2017-08-04
Fix Travis configuration for different distributions

stof:
HHVM is only available on trusty now, not on precise. And precise does not have PHP 5.3.

So this forces the distribution explicitly for both of them.
This fixes the HHVM job today, and ensures that the PHP 5.3 job keeps working when Travis migrates this repository to using trusty by default (they are currently deploying this upgrade little by little)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,6 @@ matrix:
     include:
       - php: 5.3
         dist: precise
-      - php: hhvm
-        dist: trusty
       - php: 5.6
         env: SYMFONY_VERSION='~2.7.0'
       - php: 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
 language: php
 
 php:
-    - 5.3
     - 5.4
     - 5.5
     - 5.6
     - 7.0
-    - hhvm
 
 sudo: false
 
@@ -29,6 +27,10 @@ script:
 matrix:
     fast_finish: true
     include:
+      - php: 5.3
+        dist: precise
+      - php: hhvm
+        dist: trusty
       - php: 5.6
         env: SYMFONY_VERSION='~2.7.0'
       - php: 5.6


willdurand:
Hum, hhvm fails again (https://travis-ci.org/willdurand/BazingaJsTranslationBundle/jobs/260996131). Can we mute it?

stof:
Well, I can move them to allowed_failures if you want (or remove HHVM testing entirely if you don't care debugging the job ever).
At least, now it fails with a testsuite failure, not with a Travis initialization failure.

willdurand:
Symfony dropped support for hhvm anyway, no?

stof:
yeah

willdurand:
let's remove it too then

stof:
@willdurand done

willdurand:
top, thanks!

stof:
FYI, I added a nice feature in IncenteevTranslationCheckerBundle (our bundle containing CI-friendly commands related to translations) for users of this bundle in the latest release. Missing translations used in JS files are now detected too: https://github.com/Incenteev/IncenteevTranslationCheckerBundle/releases/tag/v1.3.0

willdurand:
Cool! Maybe you should advertize the feature of your bundle into this project's README?

stof:
Note that #164 is trying to add an equivalent feature, but hooking into the Symfony extractor system rather than the Incenteev one. And this is much harder because Symfony does not let extractors decide where they extract from, and manages paths on its side instead. And JS files are often not in bundles for project-level files. This is the main reason why I created a separate interface in my bundle (and I wrote an implementation reusing Symfony extractors). 

Supporting your bundle was on my TODO list for the bundle since day 1 actually. I finally did the work.

I'm totally OK with linking to it in your readme if you agree about it.

monteiro:
👍

Issue #200, opened 2017-03-30
remove dependency on @templating service


craigh:
regarding: http://symfony.com/blog/new-in-symfony-2-7-twig-as-a-first-class-citizen

Is there a way to keep the options open as far as the universal approach of the `@templating` service but still allow a developer to remove it and only use `@twig` instead?

https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/config/services.xml#L16
https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Dumper/TranslationDumper.php#L15-L18

stof:
As you only need to render Twig templates, using Twig directly is fine IMO

craigh:
> As you only need to render Twig templates, using Twig directly is fine IMO

I do not understand your comment. My ticket points out that This bundle relies on the `@templating` service and therefore cannot use twig (only) as a first class citizen - meaning, when I use this bundle, I must call `templating` and not `twig`. I would rather do the latter if that is possible.

unless you are implying that I should override the engine service definition?

craigh:
this ticket is closed by #213. thanks @pyrech

PR #213, opened 2017-08-14
Remove templating dependency

pyrech:
This PR removes the dependency on symfony/templating component and uses Twig directly
Ref #200

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
         "symfony/finder": "~2.7|~3.1",
         "symfony/console": "~2.7|~3.1",
         "symfony/intl": "~2.7|~3.1",
-        "symfony/templating": "~2.7|~3.1",
-        "symfony/translation":  "~2.7|~3.1"
+        "symfony/translation":  "~2.7|~3.1",
+        "symfony/twig-bundle": "~2.7|~3.1"
     },
     "require-dev": {
         "symfony/asset": "~2.7|~3.1",
--- a/Controller/Controller.php
+++ b/Controller/Controller.php
@@ -4,13 +4,13 @@ namespace Bazinga\Bundle\JsTranslationBundle\Controller;
 
 use Bazinga\Bundle\JsTranslationBundle\Finder\TranslationFinder;
 use Symfony\Component\Translation\TranslatorInterface;
-use Symfony\Component\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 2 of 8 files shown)


sod:
Generating json with a template engine seems weired alltogether.

in particular:
* https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/views/config.json.twig
* https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/Resources/views/getTranslations.json.twig

monteiro:
I know that symfony wanted to use Twig by default because of the speed. @stof can you confirm this? Is it a good practice to replace the template engine for twig for speed?

pyrech:
Yep, using Twig directly instead of the whole templating component is probably a small optimization in terms of speed because of the many classes avoided.
But at first I opened this PR just because when using Flex, the bundle installation failed and forced me to install and configure the templating component even if not using it in my app

composer.json
```diff
@@ -15,8 +15,8 @@
         "symfony/finder": "~2.7|~3.1",
         "symfony/console": "~2.7|~3.1",
         "symfony/intl": "~2.7|~3.1",
-        "symfony/templating": "~2.7|~3.1",
-        "symfony/translation":  "~2.7|~3.1"
+        "symfony/translation":  "~2.7|~3.1",
+        "twig/twig": "~1.0|~2.0"
```
stof: I suggesting depending on TwigBundle instead (as this bundle actually depends on services defined in the bundle, not just on Twig)

pyrech: Done :+1:

monteiro:
@stof thanks for the clarification.

monteiro:
@pyrech let me see what are happening with the tests, since in master they failed as well.

monteiro:
@pyrech can you plz rebase with master so we can have the tests green? Thanks!

pyrech:
Done :wink:

monteiro:
@pyrech it seems that the tests are failing for PHP 5.3 :( Since in master passes, is it a problem with this PR?

pyrech:
Hum, looks like a problem with translation ordering but no idea where could it come from :confused:

monteiro:
@pyrech tested locally with php5.3 and it worked 😢 travis not.

monteiro:
@pyrech can you rebase with master. I have dropped php 5.3 support.

pyrech:
Done, sorry for the delay

monteiro:
@pyrech thanks!

PR #216, opened 2017-09-08
Test pr

monteiro:


pyrech:
Looks like tests are also failing on PHP 5.3 on #217 which is a different PR than mine. So I think there is a real problem with a specific version of PHP 5.3.
The main difference I see is that the [last successful build on master](https://travis-ci.org/willdurand/BazingaJsTranslationBundle/jobs/272882041) use PHP 5.3.29 (cli) (built: Feb 12 2015 00:55:43) when [latests PR](https://travis-ci.org/willdurand/BazingaJsTranslationBundle/jobs/273860008) use PHP 5.3.29 (cli) (built: Nov 26 2015 00:52:51)

willdurand:
Can we get rid of php 5.3 maybe?

monteiro:
@willdurand I think we should.

willdurand:
👍

monteiro:
I'm going to create a PR then. Closing this one.

PR #217, opened 2017-09-14
Uniplaces representation of duration

lcsaguiar:


stof:
This is broken. If the number is `2`, it will not select the plural anymore, as it will try to use the third alternative (and so on).

These rules are meant to stay in sync with Symfony's PluralizationRules class (as it is its port to JS)

stof:
btw, the fact that tests are not failing for this makes me think that the testsuite is incomplete.

Issue #206, opened 2017-07-12
Support dumping translations for a locale and the fallback locale in the same file


stof:
currently, when using dumped translations, we have to load 2 files: the one for the current locale, and the one for the fallback locale (otherwise fallback feature is not working fine).

It would be great if dumped translation files could include the fallback translations too, to allow loading a single file (this should be opt-in).
Note that we could optimize file size by not dumping the whole fallback translations, but only the one which are not part of the main catalogue (i.e. the ones which will fallback)

monteiro:
👍

PR #219, opened 2017-09-22
[#206] Option added to merge fallback translations

faelgabriel:
I implemented the `--merge-fallback` option on dumper command which will merge fallback translations into files which are missing translations.
Currently, it's working only with `--merge-domains` option.

This pull request deals with #206

### DIFF
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -51,6 +51,12 @@ class DumpCommand extends ContainerAwareCommand
                 null,
                 InputOption::VALUE_NONE,
                 'If set, all domains will be merged into a single file per language'
+            )
+            ->addOption(
+                'merge-fallback',
+                null,
+                InputOption::VALUE_NONE,
+                'If set, fallback translations will be merged into files which are missing translations - Currently, working only with --merge-domains option'
             );
     }
 
@@ -72,7 +78,8 @@ class DumpCommand extends ContainerAwareCommand
     {
         $formats = $input->getOption('format');
         $merge = (object) array(
-            'domains' => $input->getOption('merge-domains')
+            'domains' => $input->getOption('merge-domains'),
+            'fallback' => $input->getOption('merge-fallback')
         );
 
         if (!is_dir($dir = dirname($this->targetPath))) {
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -79,7 +79,7 @@ class DumpCommand extends ContainerAwareCommand
         $formats = $input->getOption('format');
         $merge = (object) array(
             'domains' => $input->getOption('merge-domains'),
-            'fallback' => $input->getOption('merge-fallback')
+            'fallback' => $input->getOption('merge-fallback'),
         );
 
         if (!is_dir($dir = dirname($this->targetPath))) {
--- a/Dumper/TranslationDumper.php
+++ b/Dumper/TranslationDumper.php
@@ -207,24 +207,23 @@ class TranslationDumper
         }
     }
 
-    private function dumpTranslationsPerLocale($pattern, array $formats, $target, bool $mergeFallback)
+    private function dumpTranslationsPerLocale($pattern, array $formats, $target, $mergeFallback)
     {
         $allTranslations = $this->getTranslations();
 
         foreach ($allTranslations as $locale => $domains) {
             $translations = array($locale => $domains);
 
-            if ($mergeFallback && $locale !== $this->localeFallback) {
-                foreach ($allTranslations[$this->localeFallback] as $domainIndex => $domainValue) {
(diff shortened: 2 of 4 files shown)


Command/DumpCommand.php
```diff
@@ -72,7 +78,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $formats = $input->getOption('format');
         $merge = (object) array(
-            'domains' => $input->getOption('merge-domains')
+            'domains' => $input->getOption('merge-domains'),
+            'fallback' => $input->getOption('merge-fallback')
```
stof: I suggest adding the trailing comma here, to avoid future diffs like the one done on the previous line

Dumper/TranslationDumper.php
```diff
@@ -207,14 +207,34 @@ private function dumpTranslationsPerDomain($pattern, array $formats, $target)
         }
     }
 
-    private function dumpTranslationsPerLocale($pattern, array $formats, $target)
+    private function dumpTranslationsPerLocale($pattern, array $formats, $target, bool $mergeFallback)
```
stof: if the bundle start using scalar typehints, this requires PHP 7.

This is perfectly fine with me, but only if we tell Composer about it (in composer.json). Otherwise, composer will happily install the new version of the bundle on PHP 5.6, meaning that any PHP requirement change has to be considered a very hard BC break (as users will have to handle dependencies by hand)

faelgabriel:
I am not able to fix the unit test on my machine.
The unit test on my machine is failing using the 2.6.5 release tag.

For now, I did disable the fallback merging on the unit test.

Command/DumpCommand.php
```diff
@@ -51,6 +51,12 @@ protected function configure()
                 null,
                 InputOption::VALUE_NONE,
                 'If set, all domains will be merged into a single file per language'
+            )
+            ->addOption(
+                'merge-fallback',
+                null,
+                InputOption::VALUE_NONE,
+                'If set, fallback translations will be merged into files which are missing translations - Currently, working only with --merge-domains option'
```
stof: this means it should throw an error if trying to use it without `merge-domains` instead of silently ignoring it.

Dumper/TranslationDumper.php
```diff
@@ -207,14 +207,33 @@ private function dumpTranslationsPerDomain($pattern, array $formats, $target)
...
+    private function dumpTranslationsPerLocale($pattern, array $formats, $target, $mergeFallback)
     {
-        foreach ($this->getTranslations() as $locale => $domains) {
+        $allTranslations = $this->getTranslations();
+
+        foreach ($allTranslations as $locale => $domains) {
+            $translations = array($locale => $domains);
+
+            if ($mergeFallback) {
+                $localesFallback = array_unique(array(current(explode('_', $locale)), $this->localeFallback));
```
stof: that's broken. You can have multiple `_` in a locale, which defines multiple levels of fallbacks.

Dumper/TranslationDumper.php
```diff
@@ -207,14 +207,33 @@ private function dumpTranslationsPerDomain($pattern, array $formats, $target)
...
+                foreach ($localesFallback as $localeFallback) {
+                    if (!empty($localeFallback) && $locale !== $localeFallback && isset($allTranslations[$localeFallback])) {
+                        foreach ($allTranslations[$localeFallback] as $domain => $messages) {
+                            foreach ($messages as $key => $message) {
+                                if (!isset($translations[$locale][$domain][$key])) {
+                                    $translations[$localeFallback][$domain][$key] = $message;
```
stof: if you have multiple fallback locales, the merging should be smarter:

- first fallback locale should load all keys not present in the locale (already handled here)
- further fallback locales should load only keys not present in the locale *and any previous fallback locales*

this can be handles by building a second array in parallel, and using that one for detection:

```php
$translations = array($locale => $domains);
$knowTranslations = $domains;

// Asuming proper building of the list of fallbacks
foreach ($localesFallback as $localeFallback) {
    if (!isset($allTranslations[$localeFallback])) {
        continue;
    }

    foreach ($allTranslations[$localeFallback] as $domain => $messages) {
        foreach ($messages as $key => $message) {
            if (!isset($knowTranslations[$domain][$key])) {
                $translations[$localeFallback][$domain][$key] = $message;
                $knowTranslations[$domain][$key] = $message;
            }
        }
    }
}

stof:
@faelgabriel any chance to make the requested changes ?

stof:
@faelgabriel any news here ?

faelgabriel:
@stof Not yet, I will on it ASAP.

David-Crty:
Would love to have this feature merged, very useful :+1:

monteiro:

PR #220, opened 2017-10-11
Register command manually to avoid warning in Symfony 3.4

Cydonia7:
Command auto-registration has been deprecated in Symfony 3.4, this PR aims to fix the generated warning: `Auto-registration of the command
  "Bazinga\Bundle\JsTranslationBundle\Command\DumpCommand" is deprecated since Symfony 3.4 and won't
  be supported in 4.0. Use PSR-4 based service discovery instead.`

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -21,5 +21,8 @@
             <argument></argument> <!-- active locales -->
             <argument></argument> <!-- active domains -->
         </service>
+        <service id="bazinga.jstranslation.dump_command" class="Bazinga\Bundle\JsTranslationBundle\Command\DumpCommand">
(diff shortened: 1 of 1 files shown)


Resources/config/services.xml
```diff
@@ -21,5 +21,8 @@
             <argument></argument> <!-- active locales -->
             <argument></argument> <!-- active domains -->
         </service>
+        <service id="Bazinga\Bundle\JsTranslationBundle\Command\DumpCommand">
```
stof: omitting the class name to rely on the id is not possible, as the bundle supports Symfony DI older than 3.3.

Cydonia7: Thanks, I just fixed it

PR #225, opened 2017-11-18
Made test compatible with php 7.0 and up.

yannickl88:
Due to a change in sorting order there are some keys in a different order. I added new tests and skipped the old one so it does not involve a change in the actual code. 

I do not think the key order is crucial, so adding a sort would be unnecessary overhead.

#SymfonyConHackday2017

### DIFF
--- a/Tests/Dumper/TranslationDumperTest.php
+++ b/Tests/Dumper/TranslationDumperTest.php
@@ -34,6 +34,21 @@ t.add(15, "code postal", "numerics", "en");
 
 JS;
 
+    const JS_EN_MERGED_TRANSLATIONS_PHP7 = <<<JS
+(function (t) {
+// en
+t.add("hello", "hello", "messages", "en");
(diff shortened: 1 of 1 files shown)


monteiro:
Thanks for the contribution!

yannickl88:
hmm, could be a platform thing. There is another PR which seems to be fine. I just cannot get this to work locally. 

https://travis-ci.org/willdurand/BazingaJsTranslationBundle/builds/303913975?utm_source=github_status&utm_medium=notification

monteiro:
@yannickl88 try to rebase because there was a recent PR that solved errors related with the container and symfony 4.

yannickl88:
Sorry for the late reply. It seems the other PR fixed the issue I had with the bad sorting order. Running the test on 7.1 seems to be fine. I'll close this then.

monteiro:
Awesome thanks !

PR #230, opened 2017-12-06
Handle Symfony default web path +/- v4.0

GwendolenLynch:
Symfony 4+ defaults to `public/` for the web root. If I have done my job properly this should cover that for all supported Symfony versions.

Added tests for the command too. They are a separate commit as to me it made more sense, as it wasn't an adjustment of tests for a single change, rather initial coverage for the whole SUT. Happy to squash them though :smile:

### DIFF
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -8,6 +8,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Flex\Recipe;
 
 /**
  * @author Adrien Russo <adrien.russo.qc@gmail.com>
@@ -86,7 +87,11 @@ class DumpCommand extends Command
     {
         parent::initialize($input, $output);
 
-        $this->targetPath = $input->getArgument('target') ?: sprintf('%s/../web/js', $this->kernelRootDir);
+        $this->targetPath = $input->getArgument('target') ?: sprintf(
+            '%s/../%s/js',
+            $this->kernelRootDir,
+            class_exists('Symfony\Flex\Recipe') ? 'public' : 'web'
+        );
     }
 
     /**
--- a/Tests/Command/DumpCommandTest.php
+++ b/Tests/Command/DumpCommandTest.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace Bazinga\JsTranslationBundle\Tests\Command;
+
+use Bazinga\Bundle\JsTranslationBundle\Command\DumpCommand;
+use Bazinga\Bundle\JsTranslationBundle\Tests\WebTestCase;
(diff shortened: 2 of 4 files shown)


monteiro:
@GawainLynch is this current blocking symfony 4+ users to use this, at least starting a new project?

GawainLynch:
@monteiro well, blocking them using the dump command, but not the rest of the bundle. But if it is a new porject then you won't have the translations dumped, so … :smile_cat:

Command/DumpCommand.php
```diff
@@ -86,7 +87,11 @@ protected function initialize(InputInterface $input, OutputInterface $output)
     {
         parent::initialize($input, $output);
 
-        $this->targetPath = $input->getArgument('target') ?: sprintf('%s/../web/js', $this->kernelRootDir);
+        $this->targetPath = $input->getArgument('target') ?: sprintf(
+            '%s/../%s/js',
+            $this->kernelRootDir,
+            Kernel::VERSION_ID >= 40000 ? 'public' : 'web'
```
pierredup: Not all Symfony 4 projects would use `public`, E.G if you upgrade a SF3 project to SF4 but decide to keep the `web` directory instead of changing it to `public

stof: better would be to detect the fact that this is a Flex project, using `class_exists(Recipe::class)` (with the right use statement)

Tests/Command/DumpCommandTest.php
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @covers \Bazinga\Bundle\JsTranslationBundle\Command\DumpCommand
+ *
+ * @author Gawain Lynch <gawain.lynch@gmail.com>
+ */
+class DumpCommandTest extends WebTestCase
+{
+    private $transDumper;
+
+    public function setUp()
```
stof: should be protected, not public

GawainLynch: Yup, I was just going with the flow as the one in the local `WebTestCase` was public. I've changed both.

Command/DumpCommand.php
GawainLynch: On looking at it, I am wondering if it isn't better still to do a `file_exists()`, as even if it isn't a Flex project it could still be in either. Thoughts?

stof: I would not do a file_exists, as this might create issue (for instance, if you do it on the `js` subfolder, it might not exist yet).
This is about the default behavior anyway. So anyone using a custom structure can already pass the `target` argument explicitly.

GawainLynch: I was thinking checking the web vs public, but fair point regardless. :+1:

SpartakusMd:
Hello. Can we assume that this pull request will be merged soon as it blocks the usage of the bundle on Symfony 4 ?

pierredup:
You can still use the `--target` option of thr command to set the path, so this is still usable on Symfony 4

monteiro:

PR #231, opened 2017-12-15
Tweaked Travis config: test ~3.4.0 and skip beta deps

bocharsky-bw:
We do not need to test beta deps - Symfony 4 is stable now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
     - 5.5
     - 5.6
     - 7.1
+    - 7.2
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


stof:
well, what the Travis matrix should do is testing dev deps rather than beta deps (making the job useful all year, and not only a few weeks per year).

Having a new job for the 3.4 LTS makes sense though.

bocharsky-bw:
Good point, I added testing dev deps

monteiro:
Thanks @bocharsky-bw

PR #235, opened 2017-12-15
Add TOC for convenient navigation through the docs

bocharsky-bw:


willdurand:
I would not add the full links but :+1:

bocharsky-bw:
Good point, I'll make them relative

Resources/doc/index.md
```diff
@@ -10,6 +10,22 @@ of this bundle. For version `1.x`, please read:
 Also, you might be interested in this [UPGRADE
 guide](https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/UPGRADE.md).
 
+- [Installation](/blob/master/Resources/doc/index.md#installation)
```
stof: that won't work due to the way Github handles relative links when rendering doc (it treats them as inside the repo, so `/blob/master` should not be there).

and as these are only links inside the same page, you could even simplify it using `#installation` only

bocharsky-bw: You're right, thanks!

monteiro:
👍

Issue #258, opened 2019-01-04
Symfony 4.2 depreciation error


steveWinter:
A tree builder without a root node is deprecated since Symfony 4.2 and will not be supported anymore in 5.0.

vanputten:
Fixed in #271

PR #259, opened 2019-01-04
#258 Symfony 4.2 tree builder depreciation error

steveWinter:
Addressing #258

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -17,9 +17,11 @@ class Configuration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        $builder = new TreeBuilder();
+        $builder = new TreeBuilder('bazinga_js_translation');
+        // BC for symfony/config < 4.2
+        $root = method_exists($builder, 'getRootNode') ? $builder->getRootNode() : $builder->root('bazinga_js_translation');
 
-        $builder->root('bazinga_js_translation')
+        $root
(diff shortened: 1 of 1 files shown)


steveWinter:
Open to suggestions as to what to do about the Travis failures 
- for PHP > 7 this is being caused by another Symfony 4.2 deprecation - I'm happy to address this, but would prefer to do so in another PR at a later time
- for PHP < 7 composer require is running out of memory (and given eol of 5.6 is it even necessary to support this any longer?)

monteiro:
I've created this issue: https://github.com/willdurand/BazingaJsTranslationBundle/issues/261 to the response fo the community about this. 

Thanks for the contribution!

ping-localhost:
Is there anything I can do to get this merged @monteiro?

stof:
composer running out of memory is not for PHP < 7 but for jobs customizing the constraints. The fact that they run PHP 5.6 is not the cause of the issue (we also have green jobs with PHP 5.6). So dropping support for PHP 5.6 is not the solution.

The deprecation for jobs running PHP 7 (and so installing Symfony 4) is because of the testsuite itself relying on a deprecated feature. That could indeed be solved separately.

And there is another uninvestigated failure: the PHP 5.4 job getting the wrong data (maybe this is related to using Symfony 2.8 in it rather than Symfony 3)

steveWinter:
It seems #263 is also attempting to do the same and gaining more traction so I'm going to close this.

PR #260, opened 2019-01-31
Fix deprecation "A tree builder without a root node is deprecated sin…

rumours86:
…ce Symfony 4.2 and will not be supported anymore in 5.0."

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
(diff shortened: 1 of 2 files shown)


vanputten:
This has been fixed in #271 which is merged in release 3.0.0

monteiro:
This has been already fixed. Thank you very much!

PR #267, opened 2019-11-10
fix some unit tests and increase symfony versions

monteiro:
- some tests are failing because of order (still researching why)
- increased versions in travis to include the LTS and the next symfony version 4.4 coming out soon this month.

Can you  help reviewing the file `.travis.yml` @stof ?
The goal is to support the latest LTS, plus the latest versions so we unblock the latest PRs.

*The memory problem only happens with PHP56:*
- https://stackoverflow.com/questions/51341463/unable-to-change-travis-ci-php-memory-limit/51381888 (I've put a limit, so it's not a -1)


.travis.yml
```diff
@@ -1,19 +1,20 @@
 language: php
 
 php:
-    - 5.4
-    - 5.5
     - 5.6
+    - 7.0
```
maxhelias: PHP 7.0 is no longer maintained, I think we can remove it.

Before I approve, do you want to fix all the errors in this PR or fix only the memory-limit and retry the current PRs ?

.travis.yml
```diff
@@ -1,19 +1,20 @@
 language: php
 
 php:
-    - 5.4
-    - 5.5
```
stof: this should be dropped only if the composer.json removes them. Otherwise, you should force to run them on trusty instead (which you can do by specifying them in the matrix config to configuring `dist` for these jobs)

.travis.yml
```diff
@@ -29,11 +30,9 @@ script:
...
-      - php: 5.6
-        env: SYMFONY_VERSION='~2.7.0'
-      - php: 5.6
-        env: SYMFONY_VERSION='~2.8.0'
-      - php: 5.6
-        env: SYMFONY_VERSION='~3.4.0'
-      - php: 7.2
+      - php: 7.1.3
+        env: SYMFONY_VERSION='~3.4.0'        
+      - php: 7.1.3
```
stof: why specifying a patch version ?

monteiro:
Close in favor of #271

PR #268, opened 2019-11-12
Handled a deprecation in configuration builder with respect to symfony 4

junaidbinfarooq:


junaidbinfarooq:
Already done in another PR

PR #271, opened 2019-12-06
Add support for Symfony ^4.0|^5.0 + fix various deprecations + fix travis CI

### LINKED ISSUE #258: Symfony 4.2 depreciation error

### LINKED ISSUE #261: Deprecate PHP 5.x and PHP 7.0
Following the PHP supported versions table: http://php.net/supported-versions.php

I wonder if we can deprecate the PHP 5.x versions in this bundle.

### LINKED ISSUE #264: deprecated TreeBuilder::root() method
The "Symfony\Component\Config\Definition\Builder\TreeBuilder::root()" method called for the "bazinga_js_translation" configuration is deprecated since Symfony 4.3, pass the root name to the constructor instead.

### LINKED ISSUE #265: TreeBuilder::root deprecated
Hello,
When will the fix be applied #258 to no longer have this warning with SF4.3?
Thank you ;)

### LINKED ISSUE #269: Updated the project to Symfony 5.0
Hi,
Is it possible that you make compatible this bundle for Symfony 5?
Thank you :)

### LINKED ISSUE #270: Support for Symfony 4.4 (LTS)
**Problem**
The current master branch has no support for Symfony 4.4 (which is a long time support version) since Symfony 4.4 depends on Twig 3.0. For example, the `TranslationDumper` and `Controller` fail since Twig 3.0 uses a new namespacing format for its main environment.

**Proposed solution**
Add a 4.4 branch with support for Symfony 4.4 and Twig 3.0

vanputten:
First contribution ever 👶 

- Fix Travis CI for 3.4|^4.0|^5.0
- Add support for Symfony ^4.4 and ^5.0 to bundle
- Drops support for php <5.6 (impossible to support 5.0 otherwise)
- Drops support for Symfony <3.4 (impossible to support 5.0 otherwise)

- (Backwards) compatible with Symfony ^3.4|^4.0|^5.0
- Compatible with all major twig versions

Fixes #270 
Fixes #258 
Fixes #261 
Fixes #264 
FIxes #265 
FIxes #269


composer.json
```diff
@@ -11,19 +11,19 @@
     }
     ],
     "require": {
-        "symfony/framework-bundle": "~2.7|~3.1|~4.0",
-        "symfony/finder": "~2.7|~3.1|~4.0",
-        "symfony/console": "~2.7|~3.1|~4.0",
-        "symfony/intl": "~2.7|~3.1|~4.0",
-        "symfony/translation":  "~2.7|~3.1|~4.0",
-        "symfony/twig-bundle": "~2.7|~3.1|~4.0"
+        "symfony/framework-bundle": "4.4.*",
```
monteiro: Using this we are supporting only symfony 4.4. We need to put the pipe as before.

vanputten: I changed it to: ~4.4|~5.0 

The changes I made are breaking for Symfony 4.3 and before.

composer.json
```diff
@@ -11,19 +11,19 @@
     }
     ],
     "require": {
-        "symfony/framework-bundle": "~2.7|~3.1|~4.0",
-        "symfony/finder": "~2.7|~3.1|~4.0",
-        "symfony/console": "~2.7|~3.1|~4.0",
-        "symfony/intl": "~2.7|~3.1|~4.0",
-        "symfony/translation":  "~2.7|~3.1|~4.0",
-        "symfony/twig-bundle": "~2.7|~3.1|~4.0"
+        "symfony/framework-bundle": "~4.4|~5.0",
```
maxhelias: You should use `^4.0|^5.0`. 4.0 should be compatible if we define a conflict with twig version like this : `"twig/twig": "<1.41|<2.10"`

maxhelias:
You also need to modify the travis.yml file to fit the new constraint of PHP & Symfony

vanputten:
- travis.yml has been updated to match the Symfony and PHP versions that this update should support
- composer conflict on twig 1 and 2 has been hadden
- support for symfony 4.1 has been restored

.travis.yml
```diff
@@ -29,11 +26,15 @@ script:
...
-      - php: 5.6
-        env: SYMFONY_VERSION='~2.7.0'
-      - php: 5.6
-        env: SYMFONY_VERSION='~2.8.0'
-      - php: 5.6
-        env: SYMFONY_VERSION='~3.4.0'
+      - php: 7.2
+        env: SYMFONY_VERSION='~4.1.0'
+      - php: 7.2
+        env: SYMFONY_VERSION='~4.2.0'
```
maxhelias: No need to test 4.1 and 4.2, they are no longer maintained. Just to custom stage with `SYMFONY_VERSION=^4.0` & `SYMFONY_VERSION=^5.0`

composer.json
```diff
@@ -11,21 +11,24 @@
     }
     ],
     "require": {
-        "symfony/framework-bundle": "~2.7|~3.1|~4.0",
-        "symfony/finder": "~2.7|~3.1|~4.0",
-        "symfony/console": "~2.7|~3.1|~4.0",
-        "symfony/intl": "~2.7|~3.1|~4.0",
-        "symfony/translation":  "~2.7|~3.1|~4.0",
-        "symfony/twig-bundle": "~2.7|~3.1|~4.0"
+        "symfony/framework-bundle": "^4.0|^5.0",
```
maxhelias: It's too bad to lose compatibility with 3.4 :/

Controller/Controller.php
```diff
@@ -73,7 +73,7 @@ class Controller
      */
     public function __construct(
         TranslatorInterface $translator,
-        Twig_Environment $twig,
+        Environment $twig,
```
maxhelias: I think without type hint, you can keep the 3.4 compatibility. Take a look here : https://github.com/nelmio/NelmioApiDocBundle/pull/1560/files

vanputten: Nice solution, I implemented it to restore Symfony 2 and 3 support.

Should we maintain 2.8 support? It has become unmaintained as of november 2019

vanputten:
I removed symfony <3.3 support since `kernel.root_dir` has been removed in symfony 5.0, but its replacement (`kernel.project_dir`) was introduced in Symfony 3.3.
Besides, Symfony <3.3 are no longer maintained versions.

vanputten:
- Added support for both the new `TranslatorInterface` in `Contracts/` and for the legacy `TranslatorInterface` in `Components/`
- Support for php 5.6 is incompatible with Symfony 5.0 since Symfony's `Webtestcase` requires a return typehint in the `tearDown` method (return typehints do not exist in php 5.6)

.travis.yml
```diff
@@ -29,11 +27,11 @@ script:
 matrix:
     fast_finish: true
     include:
-      - php: 5.6
-        env: SYMFONY_VERSION='~2.7.0'
-      - php: 5.6
-        env: SYMFONY_VERSION='~2.8.0'
       - php: 5.6
         env: SYMFONY_VERSION='~3.4.0'
```
maxhelias: ^3.4, should work

.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: php
 
 php:
-    - 5.4
-    - 5.5
     - 5.6
```
maxhelias: 5.6, could be dropped

composer.json
```diff
@@ -11,20 +11,20 @@
     }
     ],
     "require": {
-        "symfony/framework-bundle": "~2.7|~3.1|~4.0",
-        "symfony/finder": "~2.7|~3.1|~4.0",
-        "symfony/console": "~2.7|~3.1|~4.0",
-        "symfony/intl": "~2.7|~3.1|~4.0",
-        "symfony/translation":  "~2.7|~3.1|~4.0",
-        "symfony/twig-bundle": "~2.7|~3.1|~4.0"
+        "symfony/framework-bundle": "~3.3|~4.0|~5.0",
```
maxhelias: Use this constraint `^3.4|^4.0|^5.0`. 3.3 can be dropped

composer.json
```diff
@@ -11,20 +11,20 @@
...
-        "symfony/yaml": "~2.7|~3.1|~4.0",
-        "symfony/browser-kit": "~2.7|~3.1|~4.0",
-        "symfony/twig-bundle": "~2.7|~3.1|~4.0",
-        "symfony/phpunit-bridge": "~2.7|~3.1|~4.0",
-        "phpunit/phpunit": "^4.8|~5.7"
+        "symfony/asset": "~3.3|~4.0|~5.0",
+        "symfony/yaml": "~3.3|~4.0|~5.0",
+        "symfony/browser-kit": "~3.3|~4.0|~5.0",
+        "symfony/twig-bundle": "~3.3|~4.0|~5.0",
+        "symfony/phpunit-bridge": "~3.3|~4.0|~5.0",
```
maxhelias: And `^3.4.24|^4.0|^5.0` only for symfony/phpunit-bridge

maxhelias:
For phpunit, you can remove the hard dependencies on the composer.json and use a file like this : https://github.com/nelmio/NelmioApiDocBundle/blob/master/phpunit

composer.json
```diff
@@ -11,20 +11,21 @@
...
-        "symfony/browser-kit": "~2.7|~3.1|~4.0",
-        "symfony/twig-bundle": "~2.7|~3.1|~4.0",
-        "symfony/phpunit-bridge": "~2.7|~3.1|~4.0",
-        "phpunit/phpunit": "^4.8|~5.7"
+        "symfony/asset": "~3.4|~4.0|~5.0",
+        "symfony/yaml": "~3.4|~4.0|~5.0",
+        "symfony/browser-kit": "~3.4|~4.0|~5.0",
+        "symfony/twig-bundle": "~3.4|~4.0|~5.0",
+        "symfony/phpunit-bridge": "~3.4|~4.0|~5.0",
+        "phpunit/phpunit": "^3.4.24|^4.0|^5.0"
```
maxhelias: You are deceived of package. The constraint was good here, this one is for symfony/phpunit-bridge

vanputten: Oops, thanks

Controller/Controller.php
```diff
@@ -109,7 +119,7 @@ public function getTranslationsAction(Request $request, $domain, $_format)
         $locales = $this->getLocales($request);
 
         if (0 === count($locales)) {
-            throw new NotFoundHttpException();
+            return (new Response())->setStatusCode(Response::HTTP_NOT_FOUND);
```
maxhelias: This should be reverted. What blocker do you have ?

vanputten: `ControllerTest.php` `testGetTranslationsWithPathTraversalAttack` and `testGetTranslationsWithLocaleInjection` expect a Response with status 404, but get a NotFoundException. 

This works fine in Symfony ^4.0 and ^5.0, but in Symfony 3.4 this fails (probably because the 3.4 exception handler does not turn it into a 404 Response object, whereas the 4.0 and 5.0 handlers do).

I thought it made more sense to return an actual 404 response, since according to the `ControllerTest` that is what we expect to get anyway (line 184 and 201 of `ControllerTest.php`).

I would say either return an actual 404 response 
**OR** change the ControllerTest to test for an uncaught `NotFoundException` instead of 404 response

If you want to reproduce the error:
- Revert these changes
- Symfony 3.4 on PHP 7.2 -> run tests

vanputten:
@maxhelias any chance you could review the latest version/my response to your comments? Would be nice if we can get this merged before too many other repos start relying on my fork.

maxhelias:
@vanputten I did not have time to watch and try it.
@monteiro may have better vision than me.

monteiro:
I will review this! Sorry for being out. My second son has just born.

maxhelias:
Don't worry, I did a first review.
And congratulations 😃

vanputten:
Congratulations @monteiro ! And awesome if you could review this! 🎉

kl3sk:
Congrats 👏 @monteiro I wait for my second daughter ^^, but this bundle require your review

Controller/Controller.php
monteiro: For now I think it is ok. Symfony 3.4 will be a pain to support always so when it gets deprecated on November 2020 we can change this piece of code.

vanputten:
Thanks for the approval @monteiro !

kl3sk:
Is this PR will be merged ? Not installable yet on SF5 from packagist

vanputten:
@kl3sk no, not merged yet. Only approved. Not sure who needs to do the merge and release

kl3sk:
Bad news, @maxhelias  and/or @monteiro  are not able to do it ?

monteiro:
I can merge. I will merge it today and create a release.

Controller/Controller.php
```diff
@@ -62,25 +63,34 @@ class Controller
...
+        $translator,
+        $twig,
         TranslationFinder $translationFinder,
         $cacheDir,
         $debug          = false,
         $localeFallback = '',
         $defaultDomain  = '',
         $httpCacheTime  = 86400
     ) {
+        if (!$twig instanceof \Twig_Environment && !$twig instanceof Environment) {
```
stof: I suggest typehinting the namespaced class name and be done with it. Namespaced class are available in 1.x and 2.x already since a long time (and Symfony 3.4 requires Twig versions having them as their min Twig version anyway).

Tests/Fixtures/app/config/base_config.yml
```diff
@@ -1,14 +1,15 @@
...
     translator:    { enabled: true }
     validation:    { enabled: false }
 
 # Twig Configuration
 twig:
-    debug:            "%kernel.debug%"
-    strict_variables: "%kernel.debug%"
+    debug:                "%kernel.debug%"
+    strict_variables:     "%kernel.debug%"
+    exception_controller: null
```
stof: if you want to keep using a NotFoundHttpException in the controller without breaking tests for Symfony 4.3 and older, the solution would be to apply this `exception_controller: null` config only on Symfony 4.4+ (because here, you disable *all* error handling on older versions as they don't have the new system)

composer.json
```diff
@@ -11,20 +11,21 @@
...
-        "symfony/yaml": "~2.7|~3.1|~4.0",
-        "symfony/browser-kit": "~2.7|~3.1|~4.0",
-        "symfony/twig-bundle": "~2.7|~3.1|~4.0",
-        "symfony/phpunit-bridge": "~2.7|~3.1|~4.0",
-        "phpunit/phpunit": "^4.8|~5.7"
+        "symfony/asset": "~3.4|~4.0|~5.0",
+        "symfony/yaml": "~3.4|~4.0|~5.0",
+        "symfony/browser-kit": "~3.4|~4.0|~5.0",
+        "symfony/twig-bundle": "~3.4|~4.0|~5.0",
+        "symfony/phpunit-bridge": "^3.4.24|^4.0|^5.0",
```
stof: The latest version of the PHPUnit bridge is compatible with PHP 5.5+ anyway. No need to support so many different versions IMO.

monteiro:
@stof just applied all your suggestions. Thanks for the review.

Dumper/TranslationDumper.php
```diff
@@ -56,21 +56,25 @@ class TranslationDumper
...
-        Twig_Environment $twig,
+        $twig,
         TranslationFinder $finder,
         Filesystem $filesystem,
         $localeFallback = '',
         $defaultDomain  = '',
         array $activeLocales = array(),
         array $activeDomains = array()
     ) {
+        if (!$twig instanceof \Twig_Environment && !$twig instanceof Environment) {
```
maxhelias: So the twig namespace could be used here also

monteiro:
Now yes! I think everything is done! :)

PR #273, opened 2020-01-07
Update index.md

melkyfb:
The Square Bracket was positioned wrongly.


verschoof:
@melkyfb maybe rebase this pr with master so the test can run again?

vibby:
This is not a mistake, the value «1» is not included in the intreval.

PR #275, opened 2020-01-14
On dump file, execute() should return int

kl3sk:
Return value of "Bazinga\Bundle\JsTranslationBundle\Command\DumpCommand::execute()" must be of the type int, NULL returned.


kl3sk:
ping @monteiro

monteiro:
Looks good!

monteiro:
Ready to review and merge?

kl3sk:
Ok for me 😃

PR #282, opened 2020-03-20
Fix Composer Deprecation

blankse:
```
Deprecation Notice: Class Bazinga\Bundle\JsTranslationBundle\Tests\AppKernel located in vendor/willdurand/js-translation-bundle\Tests\Fixtures\app\AppKernel.php does not comply with psr-4 autoloading standard. It will not autoload any
more in Composer v2.0. in phar://composer.phar/src/Composer/Autoload/ClassMapGenerator.php:185
```


blankse:
@monteiro Can you review it?

monteiro:
LGTM 👍 thanks!

PR #283, opened 2020-04-01
Exclude tests and development files from the archive

stof:
This also avoids a deprecation notice from composer when generating an optimized autoloader due to the fixture app of the tests not following the PSR-4 convention configured for sources (while being in a subfolder)


monteiro:
Absolutely.

PR #287, opened 2020-06-04
Fixes deprecation of referencing controllers with single colon

LinasRam:
Fixes this deprecation message thrown in Symfony 5.1:
`User Deprecated: Since symfony/http-kernel 5.1: Referencing controllers with a single colon is deprecated. Use "bazinga.jstranslation.controller::getTranslationsAction" instead.`

### DIFF
--- a/Resources/config/routing/routing.yml
+++ b/Resources/config/routing/routing.yml
@@ -1,6 +1,9 @@
 bazinga_jstranslation_js:
     path:  /translations/{domain}.{_format}
-    defaults: { _controller: bazinga.jstranslation.controller:getTranslationsAction, domain: "messages", _format: "js" }
+    defaults:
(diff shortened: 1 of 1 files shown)


gam6itko:
You should create another file e.g `routing-sf5` with double colon for backward compatibility.

stof:
The notation with 2 colons for service controllers works in Symfony 4.1+. So dropping support for Symfony 3.4 would be a better path forward than creating a separate file IMO (the deprecation was initially added in 4.1.0 and then reverted to avoid bundles from having to provide 2 files and then having projects choose the right one).

craigh:
this can be closed. the same fix is merged [here](https://github.com/willdurand/BazingaJsTranslationBundle/commit/279916fb96fc5e187b1df50f1b66368e15d098a3)

PR #288, opened 2020-08-06
Support for ICU-based message formatting, close #256

### LINKED ISSUE #256: Support for ICU-based message formatting
Symfony 4.2 adds support for ICU-based message formatting, but using a special convention on the domain name (it will look for messages in this suffixed domain first, and format them using the ICU rules instead of the Symfony convention).

It would be great to support that for projects migrating to the new format (maybe using the JS Intl API to have part of the info, even though it does not have the high-level MessageFormatter yet)

Kocal:
~PR under development, I need to update the JS part to support ICU templates.~

I'm thinking about using [intl-messageformat](https://github.com/formatjs/formatjs/tree/main/packages/intl-messageformat) dependency, it provides a simple API to format one message (and not a whole catalogs like other librairies).
Example:
```js
const IntlMessageFormat = require('intl-messageformat');

const mf = new IntlMessageFormat.IntlMessageFormat('Hello {name}!');
mf.format({ name: 'Bob' }) // Hello bob!
```

**EDIT:** See https://github.com/willdurand/BazingaJsTranslationBundle/pull/288#issuecomment-671024601


Tests/Fixtures/app/config/base_config.yml
```diff
@@ -15,6 +15,7 @@ bazinga_js_translation:
     active_locales:
         - fr
         - en
+        - it
```
stof, line 18: why adding italian translations in the testsuite ? this seems unnecessary to me.

Util.php
```diff
@@ -0,0 +1,24 @@
...
+     * @return array
+     */
+    public static function extractCatalogueInformationFromFilename(string $filename)
+    {
+        [$domain, $locale, $extension] = explode('.', basename($filename), 3);
+
+        // Domains using ICU Message Format suffix their names with "+intl-icu".
+        // See https://symfony.com/doc/current/translation/message_format.html#using-the-icu-message-format
+        if (false !== $icuPos = strpos($domain, '+intl-icu')) {
+            $domain = substr($domain, 0, $icuPos);
```
stof, line 19: This looks wrong to me. If you remove the `+intl-icu` suffix, you loose the info that these messages use the ICU format. The JS code will need this info.

stof:
what should happen is that the domain filter accepts both `messages` and `messages+intl-icu` files when being asked to filter by the `messages` domain.

Tests/Fixtures/app/config/base_config.yml
Kocal: I added a new locale which would be _ICU-only_, but indeed I can keep `en` and `fr` locales only and add a new `messages-intl+icu.fr.yml` catalogue.

Kocal:
How Symfony handle `messages` and `messages+intl-icu` files? Is is possible to use "classic" and ICU message formats at the same time?

Util.php
Kocal: That's why I thought some minutes ago before your early review.

Maybe return a 4th value `$format`

stof:
> How Symfony handle `messages` and `messages+intl-icu` files? Is it possible to use "classic" and ICU message formats at the same time?

yes it is. The opt-in is per file thanks to the suffix in the domain. The handling in Symfony is done at the formatter level (and in a few other places like the commands dealing with catalogues with support for a domain filtering): https://github.com/symfony/symfony/blob/669b3dfd42e382c61a2f540ca518d1ccb5a6ebee/src/Symfony/Component/Translation/Translator.php#L217-L221

Kocal:
Nice, thanks.

Kocal:
I think this PR is ready to be reviewed.

The final user should have nothing to change to use translations from ICU domains, except to make library `IntlMessageFormat` available, [see documentation](https://github.com/Kocal/BazingaJsTranslationBundle/blob/GH-256/Resources/doc/index.md#installation-1).

Given the following translations files:

```yaml
# translations/messages.en.yaml
hello: Hello!

# translations/messages+intl-icu.en.yaml
hello_name: Hello {name}!
```

This HTTP call will return translations for `messages` and `messages+intl-icu` domains:
```html
<script src="{{ url('bazinga_jstranslation_js', { 'domain': 'messages' }) }}"></script>
```

And there is nothing to change with `Translator.trans()` calls:

```js
Translator.trans('hello', {}, 'messages'); // Hello!
Translator.trans('hello_name', { name: 'John' }, 'messages'); // Hello John!
```

Like [Symfony's Translator](https://github.com/symfony/symfony/blob/669b3dfd42e382c61a2f540ca518d1ccb5a6ebee/src/Symfony/Component/Translation/Translator.php#L217-L221), ICU translations have an higher priority than "classic" translations.
It means that if you have two files `messages.en.yaml` and `messages+intl-icu.en.yaml` with the same translation key, the translation from `messages+intl-icu.en.yaml` will be used. 

Note that ICU support has not been implemented in `Translator.transChoice`:
- equivalent method in Symfony has been deprecated in 4.2 and removed in 5.0 (see https://github.com/symfony/symfony/pull/28375)
- you can use `Translator.trans` with ICU messages format instead

I've also did some improvments to the `run-qunit.js` file, it now display failed tests in the CLI. Otherwise it was impossible to know which tests where failing in PhantomJS (but were working in Chrome).

What do you think? 
Thanks!

monteiro:
Sorry for the late reply. Looks good to me. I need to fix the Travis tests.

koddistortion:
Is there any chance, that this is going to be merged anytime soon?

PR #289, opened 2020-09-25
Fix error "Uncaught SyntaxError: Unexpected end of input in translati…

marlord:
…ons line 3", if file minifier is used

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "willdurand/js-translation-bundle",
+    "name": "marlord/js-translation-bundle",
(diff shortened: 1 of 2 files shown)


tourze:
I have the same problem. Is there any params can control the minifier?

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "willdurand/js-translation-bundle",
+    "name": "marlord/js-translation-bundle",
```
stof, line 2: this should not be in this PR

monteiro:

PR #290, opened 2020-10-12
Allow domain like ```messages+icu-intl``` in routing.yml

vibby:
Allow domain to contain + and - for «+icu-intl» suffixes, when using the symfony on-the-fly generated translation.

### DIFF
--- a/Resources/config/routing/routing.yml
+++ b/Resources/config/routing/routing.yml
@@ -7,4 +7,5 @@ bazinga_jstranslation_js:
         expose: true
     requirements:
         _format: js|json
-        domain: "[\\w]+"
+        domain: "[\\w+-]+"
+


stof:
I think we should rather automatically include the suffixed domain when requesting a domain. As far as usage of the translator is concerned in Symfony, the domain is still `messages`, not `messages+icu-intl` (at lower levels, they are indeed 2 domains)

vibby:
Not sure it is a good idea, but ther should be both on the same project : ```messages.yaml``` and ```messages+icu-intl.yaml```. How to ask for a one of the two ?

Kocal:
It is already the case since #288.

E.g. if you use `<script src="{{ url('bazinga_jstranslation_js', { 'domain': 'messages' }) }}"></script>`, it will returns messages from both `messages` and `messages+icu-intl` domains, see the [tests](https://github.com/willdurand/BazingaJsTranslationBundle/pull/288/files#diff-65c376fd5fcaa50e7ac05b71078fcd9155c5446cbe97118c413295e2c8d591afR39).

PR #294, opened 2020-10-24
Remove support for Symfony < 3.3 constructor signature

### LINKED ISSUE #291: Symfony 5.2-BETA2 issue with array_merge()
My tests are failing with Symfony 5.2-BETA2 because in [this line](https://github.com/willdurand/BazingaJsTranslationBundle/blob/master/DependencyInjection/Compiler/TranslationResourceFilesPass.php#L58)

        $translatorOptions = array_merge($translatorOptions, $translator->getArgument(3));

I get this error:
> array_merge(): Argument #2 is not an array

if I replace with this:

        $translatorOptions = array_merge($translatorOptions, (array) $translator->getArgument(3));

it works properly.

can this change be integrated and released?

pierredup:
This bundle dropped support for Symfony <= 3.3 in #271, so the compiler pass doesn't need to cater for the old Symfony translator signature, and can just use the new signature from 3.3+.

Fixes #291 
Replaces #293

### DIFF
--- a/DependencyInjection/Compiler/TranslationResourceFilesPass.php
+++ b/DependencyInjection/Compiler/TranslationResourceFilesPass.php
@@ -49,14 +49,9 @@ class TranslationResourceFilesPass implements CompilerPassInterface
         $translationFiles = array();
         $translator = $container->findDefinition('translator.default');
 
-        try {
-            $translatorOptions = $translator->getArgument(4);
-        } catch (OutOfBoundsException $e) {
-            $translatorOptions = array();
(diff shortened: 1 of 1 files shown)


notFloran:
@monteiro Is it possible to make a release? This PR fixes an issue that prevents upgrading to Symfony 5.2

pierredup:
@notFloran There is already a new release available at https://github.com/willdurand/BazingaJsTranslationBundle/releases/tag/4.0.0

PR #296, opened 2020-10-24
fix travis builds and Symfony deprecation

craigh:


craigh:
@Kocal @blankse @monteiro 

can someone merge this (and then #294)? I would be great to get this bundle functional...

cc @stof

craigh:
@monteiro I see you 👍 my comment, but you did not merge. Can you do that?

.travis.yml
```diff
@@ -27,8 +27,6 @@ script:
 matrix:
     fast_finish: true
     include:
-      - php: 7.2
-        env: SYMFONY_VERSION='^3.4.0'
```
pierredup, line 31: This bundle still supports Symfony 3.4, why was this removed?

craigh: I guess because I misread it. If it can be re-added and the builds/tests pass, then great

PR #300, opened 2020-10-27
Deprecation fix

craigh:


monteiro (approved):

Tests/Dumper/TranslationDumperTest.php
```diff
@@ -153,7 +154,7 @@ public function setUp()
         $container = $this->getContainer();
 
         $this->target     = sys_get_temp_dir() . '/bazinga/js-translation-bundle';
-        $this->filesystem = $container->get('filesystem');
```
monteiro, line 156: the filesystem is not in the container?

stof: not by default, at least.

craigh: ⬆️ what he said. I assume in Symfony 4 or later the `filesystem` service became private (along with many other services).

craigh:
@stof @monteiro any ideas on solving the problem in the [travis build](https://travis-ci.org/github/willdurand/BazingaJsTranslationBundle/builds/739470390?utm_source=github_status&utm_medium=notification) with Symfony 3.4? It looks like the routing definition is incompatible with Symfony 3.4 (with 2 colons). In other words, Symfony 3.4 requires the single colon notation and the Symfony 4+ requires the double colon notation. Any ideas on how to solve that?

stof:
The discussion in https://github.com/willdurand/BazingaJsTranslationBundle/pull/287#issuecomment-656181220 was precisely suggesting to drop support for Symfony 3.4 because of this deprecation.

so the actual issue is not that https://github.com/willdurand/BazingaJsTranslationBundle/pull/296 removed the Symfony 3.4 job. It is that it forgot to change the composer requirement.

craigh:
OK, I have removed support for Symfony 3.x and properly adjusted the composer file. With this PR, all the builds are successful. Please merge this and then tag a new release.

README.md
```diff
@@ -8,7 +8,7 @@ Downloads](https://poser.pugx.org/willdurand/js-translation-bundle/downloads.png
 [![Latest Stable
 Version](https://poser.pugx.org/willdurand/js-translation-bundle/v/stable.png)](https://packagist.org/packages/willdurand/js-translation-bundle)
 
-A pretty nice way to expose your Symfony2 translation messages to your client
+A pretty nice way to expose your Symfony >4.0 translation messages to your client
```
stof, line 11: I suggest changing this to `Symfony`. this sentence does not need to duplicate the compatibility info. the composer.json is the source of truth for such thing.

composer.json
```diff
@@ -12,18 +12,19 @@
     ],
     "require": {
         "php": "^7.1",
-        "symfony/framework-bundle": "~3.4|~4.0|~5.0",
-        "symfony/finder": "~3.4|~4.0|~5.0",
-        "symfony/console": "~3.4|~4.0|~5.0",
-        "symfony/intl": "~3.4|~4.0|~5.0",
-        "symfony/translation":  "~3.4|~4.0|~5.0",
-        "symfony/twig-bundle": "~3.4|~4.0|~5.0"
+        "symfony/framework-bundle": "~4.0|~5.0",
```
stof, line 15: The min version should actually be `4.1` or higher as the double-colon notation for controllers defined as services has been implemented in 4.1.

I suggest actually bumping to 4.4 as minimum, as older 4.x versions are already EOLed.

stof (changes requested):

composer.json
craigh: @monteiro ?

craigh: @monteiro what do you think should be done here?

monteiro: @stof is right. This was an old discussion on dropping 3.4 support. It will make things easier. If someone wants support, can still use the older versions.

craigh:
@monteiro this is complete and ready for merge. Following a merge, a new release would be greatly appreciated.

craigh:
ping @monteiro

monteiro:
I will do it tomorrow. Thanks!

craigh:
Thank you for merging @monteiro 

any chance you could tag a release now? thank you!

monteiro:
@craigh It's already tagged: https://github.com/willdurand/BazingaJsTranslationBundle/releases/tag/4.0.0

Thanks for the help!

PR #305, opened 2020-12-10
Fix multiple files translations for same domain

thibautselingue:
This fixes a bug introduced in 43a51cf1fa2adc4c7bb985af8c9ebc414eea04a0.

The issue is translations are overridden for every file: a same domain cannot have translations loaded from several files anymore.

By checking if the array is already set, this behavior is fixed.

### DIFF
--- a/Controller/Controller.php
+++ b/Controller/Controller.php
@@ -138,7 +138,10 @@ class Controller
 
                 foreach ($files as $filename) {
                     [$currentDomain] = Util::extractCatalogueInformationFromFilename($filename);
-                    $translations[$locale][$currentDomain] = array();
+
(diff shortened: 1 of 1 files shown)


Arioch171:
@stof Could you take a look at this PR please? As you can see [here](https://github.com/willdurand/BazingaJsTranslationBundle/commit/43a51cf1fa2adc4c7bb985af8c9ebc414eea04a0#diff-856dd4f11efe5a5d2dd3f282f8217d3026f5cb87f0614e10c16c62d0fca662b7L142) the `$translations[$locale][$domain] = array()` is now inside a loop so we lose translations from the previous file each loop.

stof (approved):

PR #308, opened 2020-12-25
github actions support

monteiro:


monteiro:
I'm going to configure github actions instead of travis. Emulate the same behavior with different symfony versions so we make this package even more stable.

monteiro:
wrong branch

PR #309, opened 2020-12-26
Github actions support (include PHP 8 tests)

monteiro:
- Removed .travis.yaml
- Use github actions with the matrix that supports at least version 7.2 (following symfony composer.json).
- Added PHP 8 tests 

Captura de ecrã 2020-12-26, às 11 19 34

Master @stof , is there any matrix change you think we must be doing in order to cover as much versions we need.


composer.json
```diff
@@ -26,7 +26,7 @@
         "symfony/browser-kit": "~4.4|~5.0",
         "symfony/twig-bundle": "~4.4|~5.0",
         "symfony/phpunit-bridge": "^5.0",
-        "phpunit/phpunit": "^4.8|~5.7|~6.5"
+        "phpunit/phpunit": "^4.8|~5.7|~6.5|~8"
```
monteiro, line 29: support PHP 8

composer.json
```diff
@@ -11,7 +11,7 @@
     }
     ],
     "require": {
-        "php": "^7.1",
+        "php": ">=7.2.5",
```
monteiro, line 14: Just putting what symfony composer.json has.

PR #311, opened 2020-12-26
escaping XML/HTML tags according to intl-messageformat documentation

shtumi:
support translation ids like
translation.id:
  {count, plural,
    one {<b>1</b> item}
    other {<b>{count}</b> items}
  }


sustmi:
I think this could be solved using `ignoreTag` option too: https://formatjs.io/docs/intl-messageformat/#intlmessageformat-constructor

sustmi:
I created a new pull request that uses the `ignoreTag` option to solve the issue: https://github.com/willdurand/BazingaJsTranslationBundle/pull/325

PR #314, opened 2021-01-12
Fix :  When domain is not defined, get_message find translations in intl-icu format but don't format it correctly

spras:
Given the following translations file:

`# translations/messages+intl-icu.en.yaml
 hello_name: Hello {name}!`

When domain is not defined, get_message find the translation in intl-icu format, but it don't format it correctly.

`Translator.trans('hello_name', { name: 'John' })`

give

`Hello {name}!`

it should give

`Hello John!`

### DIFF
--- a/Resources/js/translator.js
+++ b/Resources/js/translator.js
@@ -291,7 +291,7 @@
                 if (has_message(_locale, _domains[i], id) ||
                     has_message(nationalLocaleFallback, _domains[i], id) ||
                     has_message(localeFallback, _domains[i], id)) {
-                    _domain = _domains[i];
+                    _domain = _domains[i].replace(INTL_DOMAIN_SUFFIX, '');
 
                     break;
                 }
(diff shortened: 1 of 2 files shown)


stof:
the fix is incomplete as it does not apply the suffix in `has_message` checks

spras:
Fix #315

spras:
> the fix is incomplete as it does not apply the suffix in `has_message` checks

@stof : the suffix is applied just after,  line 301 https://github.com/willdurand/BazingaJsTranslationBundle/pull/314/files#diff-9608ce90f4a6423da5d94bd649469c278db7dba58fee60bf7654afb1c3451e4fR301

When the trad was found in an intl-icu domain line 294, for example `messages+intl-icu` , then it was searched with twice the suffix :   `messages+intl-icu+intl-icu` , not found

spras:
Sorry, my initial description of the issue, and my last comment were not clear enough.
The `get_message` found the translation, but it didn't apply the correct format.
I added some tests cases.

monteiro:
Thanks @spras ,
Looks good to me.

monteiro (approved):

PR #317, opened 2021-02-03
Update getTranslations.js.twig

AngelikLy:
The filter "raw" is necessary to encode language with subtags in javascript file, otherwise there is an error on locale code ("en\u002DUS" for example).


monteiro (approved):

monteiro:
Hi @AngelikLy,

can you rebase and push, so we can see the green tests? 
I have fixed one issue with tests here: https://github.com/willdurand/BazingaJsTranslationBundle/pull/321

PR #320, opened 2021-07-02
Prevent translation files overwriting files from addResource calls

carakas:
array_merge with replace the translations that are in there while array_merge_recursive will append them. That way you don't loose translation sources.


carakas:
@monteiro Hi, I'm the maintainer of [Fork CMS](https://github.com/forkcms/forkcms).
I'm looking at using this bundle in our upcoming release but noticed that the translation files that we register get overwritten.
Since this is blocking for us I was wondering if it would be possible to merge and release this fix fast.
If it wouldn't be possible please let me know as well so I can look for an alternative

Kind regards

Carakas

monteiro:
Thanks for the PR @carakas,
Tests are failing for some reason.

monteiro:
@carakas can you rebase, so we can see if all tests are green? Thanks!

monteiro:
It's green. I will release next week if everything is fine.

Issue #312, opened 2021-01-03
Messages in ICU format not dumped


sweedheart:
I have a "messages+intl-icu.en.yaml" file in default translation folder.
When running the dump command, the messages domain is not dumped at all. When I remove the "+intl-icu" extension from the file-name, the domain ist dumped, but is obviously not working due to the ICU format used inside the file.

Probably caused by a continue in TranslationDumper:
`// e.g.: skip "messages+intl-icu" if "messages" exists. They will get merged after.`

As there is nothing done afterwards, the skipped domains won´t get dumped. Seems like dumping ICU translations are only working when there are already other translations in the same non-ICU domain.

validators domain ist dumped correctly when using ICU format, as there are already translations in some bundle for that domain.

aheidelberg:
I fixed this by adding another empty `messages.yaml` file without the `+intl-icu` suffix. Maybe this works for you as well.

Schyzophrenic:
I am running into the same issue, is there any way to solve this?

tpatartmajeur:
Hi, any update regarding this issue?

pacproduct:
Upgrading from v3 to v4 broke this as well, on our project.

It seems like the following line introduced the breaking change: https://github.com/willdurand/BazingaJsTranslationBundle/commit/43a51cf1fa2adc4c7bb985af8c9ebc414eea04a0#diff-16e86f5cd01bc16255a4735f7ab343d7999c1659cf9f1b7a0bee11c873c3d3c6R261 (File Dumper/TranslationDumper.php line 261)

Because it's now the "cleaned" domain that is looked up in "$activeDomains", my configuration entry "mass_updates+intl-icu" (in `bazinga_js_translation.yaml::bazinga_js_translation.active_domains`) doesn't get picked up anymore.

So I thought maybe the right approach now is to add the "cleaned up" entry in the `bazinga_js_translation.yaml` configuration file: "mass_updates" only. But if doing so makes the domain to be picked up by `TranslationDumper::getTranslations`, it still get removed in the following section by `TranslationDumper::dumpTranslationsPerDomain`:
```php
                if ($domain !== $cleanedDomain && !in_array($cleanedDomain, $domains, true)) {
                    // e.g.: skip "messages+intl-icu" if "messages" exists. They will get merged after.
                    continue;
                }
```

Considering what the comment says: "skip "messages+intl-icu" if "messages" exists", I'm wondering if the if test shouldn't be as follows instead:
```php
[...]

PR #322, opened 2021-07-13
Ensure that the ICU files are built if there are no corresponding non-ICU translations

steveWinter:
In #312 it is observed that ICU translations are not dumped at all. One suggested solution to this is to add a corresponding non-ICU file in the same domain, and while this works it's not ideal.

I believe that the check in the dumper is incorrect and that the icu file should be skipped only if the corresponding non-ICU file does exist, whereas it's currently being skipped if it doesn't exist!


steveWinter:
@pacproduct looking at your other comments in #312 I suspect you have other problems here because either this patch, or adding in the empty file as others have suggested solved the issue for me and I now have several production instances running with my fork while I wait for this PR to be merged (or the issue to be solved in some other way)

pacproduct:
You may well be right that I'm facing an issue in a particular context making it more difficult to solve ^^'

This said, considering the code being modified:
```php
            foreach ($domains as $domain => $translations) {

                $cleanedDomain = Util::cleanDomain($domain);
                if ($domain !== $cleanedDomain && in_array($cleanedDomain, $domains, true)) {
                    // e.g.: skip "messages+intl-icu" if "messages" exists. They will get merged after.
                    continue;
                }

                // [...]
```

My understanding if that we loop over `$domains` and extract the key `$domain` on one side, and `$translations` on the other side: Which makes me think that the `in_array` used above searches `$cleanedDomain` among _translations_, and not domains.
That's probably always going to return `FALSE` and avoid the loop from being broken, which in turn includes all domains (and that solves the problem - and could be breaking what this code was supposed to do initially, which is unclear to me).

What I'm trying to say I suppose is that the change suggested looks to me like it's never going to enter the `if` condition anymore.

Or am I missing something? 🤔

monteiro:

steveWinter:
I don't understand why this PR has just been summarily closed? As was commented [on the issue relatively recently](https://github.com/willdurand/BazingaJsTranslationBundle/issues/312#issuecomment-2127413438), this is still an issue in v6 - is there something wrong with my suggested fix (which I've been using in myriad projects via a fork for years)?

PR #328, opened 2021-12-14
TranslationDumper consider now <domain>+intl-icu and <domain> as the same domain

link911:
Actually `TranslationDumper::getTranslations()` called from `DumpCommand` consider `<domain>+intl-icu` and `<domain>` as two different domains.
I think that should not be the case.
Context: i have some `<domain>.<lang>.xlf` mixed to `<domain>+intl-icu.<lang>.xlf` due to removing transChoice to upgrade symfony version from 3.4 to 4.4 so no more transChoice and welcome intl-icu.
Maybe related to issue #312 ?
(First PR for me :) )

### DIFF
--- a/Dumper/TranslationDumper.php
+++ b/Dumper/TranslationDumper.php
@@ -266,17 +266,17 @@ class TranslationDumper
                 $translations[$locale] = array();
             }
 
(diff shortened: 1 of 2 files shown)


monteiro:

PR #333, opened 2022-01-30
Fix CI badge in README file

willdurand:
Hey, I just noticed the "CI badge" is still Travis CI but GitHub Actions are used now. /cc @monteiro


monteiro:
Thanks a lot @willdurand !

Issue #334, opened 2022-04-04
New form of french pluralization


nkampa:
Reference: https://unicode-org.github.io/cldr-staging/charts/38/supplemental/language_plural_rules.html#rules

We’re making the change to accommodate the “new” plural form for French. The form itself isn’t new to the language, it just hasn’t been used on the Internet until recently. The Unicode Consortium (the organization which sets standards for software internationalization) made the change in the rules and recommends supporting this added plural form. 

```xml
<source>{0} 0 days|{1} 1 day|]1,Inf] %1$s% days</source>
<target state="final">%1$s% jour| %1$s% de jours| %1$s% jours</target>
```

The second form `de` is only for numbers above `1000000`.

It does not appear to be supported in the latest versions of this library.

For the javascript library it looks like if we were to add `return number % 1000000 === 0 ? 1 : (number < 2 ? 0 : 2);` to the French switch case https://github.com/willdurand/BazingaJsTranslationBundle/blob/d23f7ff6f1992adf80f859bfd39869682fb2c63b/Resources/js/translator.js#L552
it is handled correctly.

Are there plans to bring these rule changes into this library?

PR #335, opened 2022-04-14
Fix french pluralization

nkampa:
References:
- https://github.com/willdurand/BazingaJsTranslationBundle/issues/334
- https://unicode-org.github.io/cldr-staging/charts/38/supplemental/language_plural_rules.html#rules

Accommodate the “new” plural form for French. The form itself isn’t new to the language, it just hasn’t been used on the Internet until recently. The Unicode Consortium (the organization which sets standards for software internationalization) made the change in the rules and recommends supporting this added plural form. 

```xml
<source>{0} 0 days|{1} 1 day|]1,Inf] %1$s% days</source>
<target state="final">%1$s% jour| %1$s% de jours| %1$s% jours</target>
```

The second form `de` is only for numbers above `1000000`.

### DIFF
--- a/Resources/js/translator.js
+++ b/Resources/js/translator.js
@@ -550,6 +550,8 @@
             case 'bh':
             case 'fil':
             case 'fr':
+                return number % 1000000 === 0 ? 1 : (number < 2 ? 0 : 2);
+
             case 'gun':
             case 'hi':
             case 'ln':


stof:
your patch does not apply it to French, but to a bunch of other locales too.

Also, the goal of this bundle is to reuse the Symfony translation files in JS easily. So this pluralization change should be adopted in Symfony first.

PR #338, opened 2022-07-28
[WIP] fix: ensure fallback locale is always loaded

### LINKED ISSUE #307: Fallback translatations not loaded
Hello, 

I'm currently having some trouble with the fallback behaviour.
In this method, shouldn't the fallbackLocale be added to the default locales ?
https://github.com/willdurand/BazingaJsTranslationBundle/blob/26d37937a5b074b410eee18e82ebd4ec95cdb569/Controller/Controller.php#L194

Because, currently it will use the explicitly specified locales or the request locale .
In every cases, Translator fallback is not working cause fallbackLocale is just not included in the return of route 'bazinga_jstranslation_js'.

As a workaround I have to include the locale as mentioned here : 
https://github.com/willdurand/BazingaJsTranslationBundle/issues/214#issuecomment-374970645

Example:
Default locale : fr
Locale : it
Fallback : en

User was by default in FR, he switched in IT but there's a missing translation key 'app.country'. It should be loaded from EN translations but it's not.

EmilePerron:
In the current version, when a fallback locale is defined, it isn't loaded automatically (as explained in #307). 

That means you still have to specifically request that the fallback be loaded, which is inconvenient and often forgotten.

This commit updates the `getLocales()` method to ensure the fallback translations are always loaded when a fallback locale is defined.

Here's a comparison in the case of a project with the `fr_CA` as the default locale and `fr` as the fallback locale (with only one message, in order to keep things simple).

**Before:**
```js
(function (t) {
t.fallback = 'fr';
t.defaultDomain = 'messages';
// fr_CA
})(Translator);
```

**After:**
```js
(function (t) {
t.fallback = 'fr';
t.defaultDomain = 'messages';
// fr_CA
// fr
t.add("generic.save", "Enregistrer", "messages", "fr");
})(Translator);
```

Closes #307

### DIFF
--- a/Controller/Controller.php
+++ b/Controller/Controller.php
@@ -210,6 +210,10 @@ class Controller
             return trim($locale);
         }, $locales));
 
+		if ($this->localeFallback && !in_array($this->localeFallback, $locales)) {
+			$locales[] = $this->localeFallback;
+		}
+
         return $locales;
     }
 }
--- a/Resources/doc/index.md
+++ b/Resources/doc/index.md
@@ -442,6 +442,9 @@ bazinga_js_translation:
                          # translator fallback.
 ```
 
+When a fallback locale is defined, that locale's translations will always
+be loaded.
+
 #### Default Domain
 
 You can define the default domain used when translation messages are added


EmilePerron:
The more I think about it, the more I think this isn't the package's responsibility.
The package has no way of knowing if the fallback locale has already been loaded in a different script or not. 

Going to close this issue - sorry for the disturbance!

PR #343, opened 2023-06-29
Fix deprecations of Symfony 6.3

stof:
This uses `@return` to opt-out of the deprecation, as done for a bunch of existing ones coming from older Symfony version.

Note that you will need to do a major version (as those classes are neither final nor internal) before being able to support Symfony 7.


monteiro:
Thanks a lot @stof !

stof:
@monteiro could you tag a new release to ship this ?

PR #345, opened 2023-07-31
Remove 'app' from DumpCommand path

mfadul24:
app path does not longer exists since sf4

### DIFF
--- a/Command/DumpCommand.php
+++ b/Command/DumpCommand.php
@@ -23,7 +23,7 @@ class DumpCommand extends Command
     /**
      * @var string
      */
-    private $kernelRootDir;
+    private $projectDir;
 
     /**
      * @var string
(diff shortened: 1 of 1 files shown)


Command/DumpCommand.php
```diff
@@ -37,7 +37,7 @@ class DumpCommand extends Command
     public function __construct(TranslationDumper $dumper, $projectDir)
     {
         $this->dumper = $dumper;
-        $this->kernelRootDir = $projectDir . '/app';
+        $this->kernelRootDir = $projectDir;
```
stof, line 40: the property should be renamed to `projectDir` instead of `kernelRootDir` then.

mfadul24:
@stof Any chance of merging this PR? Should I update anything?

stof:
@mfadul24 I don't have merge rights on this repo.

Issue #277, opened 2020-01-29
Don't use TranslatorFinder but use Translator::getCatalogue()


spiroski:
I have decorated the translator service in my project with runtime translations, but the changes from there are not reflected in the output from this bundle.

When investigating the controller, I found out that instead of using the translator's services catalogue, the bundle has it's own TranslatorFinder which finds the files, then loads them through the loader to create the catalogue:

```php 
           foreach ($locales as $locale) {
                $translations[$locale] = array();

                $files = $this->translationFinder->get($domain, $locale);

                if (1 > count($files)) {
                    continue;
                }

                $translations[$locale][$domain] = array();

                foreach ($files as $filename) {
                    $extension = pathinfo($filename, \PATHINFO_EXTENSION);

                    if (isset($this->loaders[$extension])) {
                        $resources[] = new FileResource($filename);
                        $catalogue   = $this->loaders[$extension]
                            ->load($filename, $locale, $domain);

                        $translations[$locale][$domain] = array_replace_recursive(
                            $translations[$locale][$domain],
                            $catalogue->all($domain)
                        );
                    }
                }
            }
```

Wouldn't it be better and more elegant to get the catalogue from the translator service? It would also work in cases like mine where I'm loading translations into the catalog at runtime, and I don't have a loader.

I can implement this and submit a PR if there is no reason to not do it that way.

kl3sk:
Sound a good idea, it may add (I said may) a BC break, but indeed a good idea.

biegacz1:
Any progress on this one? It would be useful change because it would allow clearing cache of translation while using https://github.com/lexik
/
[LexikTranslationBundle](https://github.com/lexik/LexikTranslationBundle)

PR #346, opened 2023-10-19
Load translations from Translator's message catalogue

drola:
Load translations from Translator's message catalogue

**Context:**

Existing implementation of BazingaJsTranslationBundle re-implements loading messages from files on disk. However, this implementation is not 1-to-1 compatible with message loading in Symfony's translator component. Thus some translations are not visible to BazingaJsTranslationBundle.

For example, [LexikTranslationBundle](https://github.com/lexik/LexikTranslationBundle) stores messages in a DB,

This PR re-implements message discovery, such that it relies on `Symfony\Component\Translation\Translator`.

This change was suggested in #277.

**Backwards Compatibility**:
 - APIs, nor output of Dumper or Controller didn't change.
 - TranslationFinder does not exist anymore.


Util.php
```diff
@@ -20,4 +16,33 @@ public static function cleanDomain(string $domain): string
...
         return $domain;
     }
+
+    /**
+     * Retrieve messages from translator bag.
+     *
+     * @param TranslatorBagInterface $translatorBag
+     * @param string $locale
+     * @param string $domain
+     * @return array 2D array, where first-level keys are domain without suffixes, and domain with '+intl-icu' suffix. Second level keys are translation keys.
```
stof, line 26: `@return array<string, array<string, string>>` would allow IDEs and static analyzers to known how this array looks like instead of having only a human-readable explanation (which is still needed to understand what the strings are about)

drola: Thank you! I've incorporated the more precise type annotation.

monteiro:
Hi @drola , can you rebase it? Thanks for your first contribution!

drola:
> Hi @drola , can you rebase it? Thanks for your first contribution!

Hey, @monteiro! It's rebased now.

Happy to help, I use this bundle in several projects.

betoobe6:
Hi there,

I'm glad someone solved this ! Do you know when it will be available in the release ?
We are using LexikTranslationBundle and would love to have only one translation WF !

Thanks

monteiro:

PR #348, opened 2023-12-14
Add support for Symfony 7

### LINKED ISSUE #347: Support Symfony 7
Symfony 7.0 was released on 29 Nov 2023. This bundle should support it.

stof:
Closes #347 

Due to the requirement to add the native return type in the Configuration class and the DumpCommand class which were neither final nor internal, this should be released as a new major version to respect semver.
All the `void` return type I'm adding as native return types are actually optional in Symfony 7. They will become required only in Symfony 8. As adding them also requires a major version, I'm adding them at the same time (so that Symfony 8 does not require a new bump of the major version).

I also took the opportunity to mark classes as `@final` or `@internal` to reduce the BC API of the bundle for the future. Note that for now, the command and controller are only marked as `@final` but I'm wondering whether they could be marked as `@internal` instead if they are only meant to be initialized by the service definition of this bundle.

### DIFF
--- a/BazingaJsTranslationBundle.php
+++ b/BazingaJsTranslationBundle.php
@@ -9,13 +9,12 @@ use Bazinga\Bundle\JsTranslationBundle\DependencyInjection\Compiler\AddLoadersPa
 
 /**
  * @author William DURAND <william.durand1@gmail.com>
+ *
+ * @final
  */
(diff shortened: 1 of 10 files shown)


monteiro:
Today or tomorrow max.

PR #350, opened 2023-12-18
Fix the branch alias for the master branch

stof:
In #348, I bumped the branch alias for a new major version. But it was actually far outdated. This bumps it to the actual next major version.


stof:
I also made the branch alias in a way that needs a bump only for major versions, not for minor ones (actually, I already did that in my previous PR)

PR #362, opened 2024-07-04
Use the UMD build of intl-messageformat in the testsuite

stof:
The main file of the library is a CommonJS build, not an UMD build, making it incompatible with the usage in the testsuite setup. The UMD build is shipped in the package but is not exposed on CDNJS.


stof:
Duplicate submission of #361

PR #363, opened 2024-07-04
Fix the loading of the Intl JS polyfills

stof:
The Intl polyfill has been split into dedicated polyfills in 2020.
As the Cloudflare service allows to specify a version of the polyfill-library, the version is now locked to avoid being affected by such change again.


stof:
hmm, apparently, this is not enough. It looks like the polyfill service decides that Intl is supported by the browser. Maybe phantomjs is not detected properly.

stof:
Given that phantomjs is dead, I will submit a PR replacing the JS test runner instead (which might remove the need for polyfills entirely thanks to having native Intl support)

stof:
Closing in favor of #367

PR #368, opened 2024-10-21
Clean the CI setup

stof:
- Remove jobs allowed to fail for stable PHP versions (keeping only jobs that requires passing, making that the default)
- Add CI jobs for PHP 8.2, 8.3 and 8.4
- Add an experimental (i.e. allowed to fail) job for PHP nightly
- Remove the `variant` from the matrix as it is always `normal`
- Remove the step fetching the composer cache location as this location is never used
- Upgrade the `ramsey/composer-install` action to avoid deprecation warnings
- ensure we don't cancel jobs for other PHP versions when one fails, to make it easier to determine whether a failure is version-specific.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -19,27 +19,16 @@ jobs:
         continue-on-error: ${{ matrix.allowed-to-fail }}
 
         strategy:
+            fail-fast: false
             matrix:
-                php-version: ['7.4', '8.0', '8.1']
-                variant: [normal]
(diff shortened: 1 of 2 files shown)


stof:
failures with Symfony 7 are being handled in #369

stof:
@monteiro which solution do you prefer ?

stof:
@monteiro I rebased this PR after the merge of my other PRs (as they were conflicting) so it is now ready to be reviewed as well.

monteiro (approved):

PR #369, opened 2024-10-21
Update WebTestCase to be compatible with Symfony 7

stof:


stof:
@monteiro any preference for the solution ?

monteiro:
I think we can bump the version.

stof:
@monteiro I updated the PR to bump the min PHP version.

monteiro:
thanks!

Issue #336, opened 2022-04-28
intl-messageformat dependency missing


mcorteel-harel:
The javascript module (published [here](https://www.npmjs.com/package/bazinga-translator)) requires the `intl-messageformat` library, but it isn't listed as a dependency in the `package.json`.
As a result, when installed using yarn or npm it doesn't work unless the user adds the dependency himself (with the risk of version issues).
Could you add the right dependency to the `package.json`?

Thanks for this great library!

PR #379, opened 2026-04-09
Improve the npm package metadata

stof:
- add metadata that allows npmjs.com to link back to the repository
- add an optional peer dependency on `intl-messageformat` so that tools being strict about dependency access (pnpm, Yarn PnP mode) can be supported (relates to https://github.com/willdurand/BazingaJsTranslationBundle/issues/336)

To make the package even better, it might make sense to create a `README.md`, as npmjs.com will render it.

Regarding the `intl-messageformat`, I'm actually wondering whether it should be defined as a required dependency instead (and then probably a normal dependency, not a peer one) as the way the UMD wrapper is defined does not make it optional when using CommonJS or RequireJS (the `require()` is always done):
https://github.com/willdurand/BazingaJsTranslationBundle/blob/6fbd01ddc4c73957e6776aa0ca1ba32e20ca889b/Resources/js/translator.js#L5-L18

### DIFF
--- a/Resources/package.json
+++ b/Resources/package.json
@@ -7,6 +7,10 @@
     "symfony"
   ],
   "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/willdurand/BazingaJsTranslationBundle.git"
+  },
   "author": {
     "name": "William Durand",
     "email": "will+git@drnd.me"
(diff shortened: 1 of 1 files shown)


stof:
I used the existing dev dependency on `intl-messageformat` 10.x for the peer dependency constraint. The latest version of the package is version 11, which is now shipping in ESM format.

PR #380, opened 2026-05-05
feat: Use Symfony cache warmer

trandbert37:


Resources/doc/index.md
```diff
@@ -377,6 +378,32 @@ By default, the `locale` is set to the value defined in the `lang` attribute of
 the `html` tag.
 
 
+Cache Warmup
```
stof, line 381: This feature should be configurable IMO.

In productions, I'm dumping translations to static files (thanks to the dump command of the bundle), which means that the controller is never used, and warming up its cache during deployment is wasting resources.
And a project is probably not using both the JS and JSON formats, so even when using the controller, it might make sense to warm up the cache only for the relevant format.

Controller/Controller.php
```diff
@@ -122,11 +139,78 @@ public function getTranslationsAction(Request $request, $domain, $_format)
...
+    {
+        return true;
+    }
+
+    /**
+     * @param string $domain
+     * @param array  $locales
+     * @param string $format
+     *
+     * @return ConfigCache
```
stof, line 205: Returning a ConfigCache is a weird API. A more sensible API would be `generateContent(string $domain, array $locales, string $format): string` (name to be bikeshed), where you return the content of the resource (i.e. returning `file_get_contents($cachePath)`), keeping the usage of the ConfigCache encapsulated inside the method.

Controller/Controller.php
```diff
@@ -122,11 +139,78 @@ public function getTranslationsAction(Request $request, $domain, $_format)
...
+            return array();
+        }
+
+        if (empty($domains)) {
+            $domains = array($this->defaultDomain ?: 'messages');
+        }
+
+        foreach ($domains as $domain) {
+            foreach ($locales as $locale) {
+                $this->generateCache($domain, (array) $locale, 'js');
```
stof, line 184: You should document that this warms the cache only for responses involving a single locale (while the controller supports requesting multiple locales in a single request, which will still be cache misses on the first request, defeating the purpose of the warmup in projects relying on that feature)

